PR #5, opened 2019-02-12
Fmu

kbenne:


worker/addfmusite/addFMUSite.py
```diff
@@ -0,0 +1,99 @@
...
+        return tarinfo
+
+    tarname = "%s.tar.gz" % site_ref
+    tar = tarfile.open(tarname, "w:gz")
+    tar.add(directory, filter=reset, arcname=site_ref)
+    tar.close()
+
+    bucket.upload_file(tarname, "parsed/%s" % tarname)
+    
+    #os.remove(tarname)
```
kbenne: @goldcoder01  since we are uploading back to s3, go ahead and remove all of the files so that the worker does not fill up

worker/addfmusite/addFMUSite.py
```diff
@@ -0,0 +1,99 @@
...
+
+shutil.rmtree(directory)
```
kbenne: Oh. never mind. this line probably gets things cleaned up. Might want to double check that there are no remnants.

worker/addfmusite/addFMUSite.py
```diff
@@ -0,0 +1,99 @@
...
+# fmu files gets uploaded with user defined names, but here we rename
+# to model.fmu to avoid keeping track of the (unreliable, non unique) user upload name
+# createFMUTags will however use the original fmu upload name for the display name of the site
+fmupath = os.path.join(directory, 'model.fmu')
+jsonpath = os.path.join(directory, 'tags.json')
+
+bucket = s3.Bucket('alfalfa')
+bucket.download_file(key, fmupath)
+
+call(['python', 'addfmusite/createFMUTags.py', fmupath, fmu_upload_name, jsonpath])
```
kbenne: All of the code before and after this line, is very similar to addSite.py which corresponds to the tagging for osm files. I wonder if we can factor this out to a common directory, and distinguish only by the create<FMU|osm>Tags.py that is called.

worker/addfmusite/createFMUTags.py
```diff
@@ -0,0 +1,72 @@
...
+fmupath = sys.argv[1]
+fmu_upload_name = sys.argv[2]
+jsonpath = sys.argv[3]
+fmu = load_fmu(fmupath)
+
+#2.0 get input/output variables from the FMU
+input_names = fmu.get_model_variables(causality = 2).keys()
+output_names = fmu.get_model_variables(causality = 3).keys()
+
+#print ("input names are: )))))): ", input_names )
```
kbenne: I suggest go ahead and remove this comments.

worker/addfmusite/createFMUTags.py
```diff
@@ -0,0 +1,72 @@
...
+    "simType": "s:fmu",
+    "siteRef": "r:%s" % siteid
+}
+tags.append(sitetag)
+
+#4.0 add input tagging
+for var_input in input_names:
+    tag_input = {
+        "id": "r:%s" % uuid.uuid4(),
+        "dis": "s:%s" % var_input+':input',
```
kbenne: I think we should avoid adding ':input' and instead make the display name simply s:<var_input>. We can use a haystack tag like :writable to signal that it is an input.

worker/runfmusite/runFMUSite.py
```diff
@@ -0,0 +1,226 @@
...
+from dateutil.parser import parse
+from pyfmi import load_fmu
+import copy
+import testcase
+import json
+
+
+def get_tag_data(tag_filepath):
+    '''
+     Purpose: retrieving the haystack tagged data,
```
kbenne: Love the comments in all of these functions.

worker/runfmusite/runFMUSite.py
```diff
@@ -0,0 +1,226 @@
...
+             a dictionary matching all outputs and IDs 
+    '''
+    dis_and_ID={}
+    inputs_and_ID = {}
+    outputs_and_ID = {}
+
+    tag_data = get_tag_data(tag_filepath)
+
+    for point in tag_data:
+        var_name = point['dis']
```
kbenne: Note that 'dis' will have :input and :output appended and that will not match the fmu keys.

worker/runfmusite/runFMUSite.py
```diff
@@ -0,0 +1,226 @@
...
+ 
+    #run the FMU simulation
+    kstep=0 
+    #while tc.start_time <= 1000000000000:
+    for i in range(10):
+        time.sleep(5)
+        tc.advance(u)
+        output = tc.get_results()
+        
+        u_output = output['u']
```
kbenne: u is input isn't it?

worker/runfmusite/runFMUSite.py
```diff
@@ -0,0 +1,226 @@
...
+        y_output = output['y']
+        for key in u_output.keys():
+            value_u = u_output[key]
```
kbenne: I think there is some work to do on u (inputs). we should retrieve the current user requested input value (if any) from the database "write_arrays" document. the set those values on the u array before calling advance.

worker/runfmusite/runFMUSite.py
```diff
@@ -0,0 +1,226 @@
...
+                cur_value = 3.33
+                recs.update_one( {"_id": input_id }, {"$set": {"rec.curVal":"n:%s" %cur_value, "rec.curStatus":"s:ok","rec.cur": "m:" }} )
+
+        for key in y_output.keys():
+            value_y = y_output[key]
+            
+            if key!='time': 
+                output_id = tagid_and_outputs[key]
+                             
+                cur_value =273.15
```
kbenne: output handling looks good. we just need to use value_y instead of the hard coded cur_value

worker/runfmusite/runFMUSite.py
```diff
@@ -0,0 +1,226 @@
...
+            if key!='time':
+                input_id = tagid_and_inputs[key]
+                                
+                cur_value = 3.33
+                recs.update_one( {"_id": input_id }, {"$set": {"rec.curVal":"n:%s" %cur_value, "rec.curStatus":"s:ok","rec.cur": "m:" }} )
+
+        for key in y_output.keys():
+            value_y = y_output[key]
+            
+            if key!='time':
```
kbenne: time format is a little different for the FMUs, but nevertheless we should report the current elapsed time in seconds, similar to how we push time for energyplus simulations. Refer to here https://github.com/NREL/alfalfa/blob/fmu/worker/runsite/runSite.py#L609

worker/runfmusite/testcase.py
```diff
@@ -0,0 +1,233 @@
+# -*- coding: utf-8 -*-
```
kbenne: this file is added to the repository in runfmusite and addfmusite. factor it out to a common directory worker/common/.

worker/runfmusite/testcase.py
```diff
@@ -0,0 +1,233 @@
...
+"""
```
kbenne: Also in this version we are missing the appropriate copyright to IBPSA.

worker/addfmusite/addFMUSite.py
goldcoder01: yes, the shutil.rmtree() cleans all the files

worker/addfmusite/createFMUTags.py
goldcoder01: resolved.

worker/addfmusite/createFMUTags.py
goldcoder01: Thanks. Initially i used 'input'/'output' to distinguish the input_vars and output_vars. This is resolved by checking if "writable" tagging property existed in the haystack vars, which distinguishes the variables.

worker/runfmusite/runFMUSite.py
goldcoder01: yes. I hard-coded the y to make sure the values were sent to the database and shown in the front-end. Updated results are pushed back to repo.

worker/runfmusite/runFMUSite.py
goldcoder01: yes, it is the input from user. But it is a trajectory with all historical values.

worker/runfmusite/runFMUSite.py
goldcoder01: resolved.

worker/runfmusite/testcase.py
goldcoder01: added.

worker/runfmusite/runFMUSite.py
goldcoder01: I see you mean: user-specify-inputs --->send-inputs-to-database ---> query-database-for-inputs --->send-queried-inputs-to-frontend?

goldcoder01: i used the "mongodb.find_one(tag_id)" to implement the same idea.

worker/runfmusite/runFMUSite.py
goldcoder01: i revised the Sites.js to report the elapsed time.

worker/runfmusite/testcase.py
goldcoder01: finished.

worker/addfmusite/addFMUSite.py
goldcoder01: This one is done!

PR #6, opened 2019-02-27
local test is done, hard-coded

YanfeiNREL:
This branch is good, for making local test through shared volume between docker and host machine.
I suggest to leave it there for future purpose.


nllong:
an official testing framework has been added. If this is still needed, then migrate to new framework.

PR #15, opened 2019-04-18
deployment scripts and docker-compose 3

anyaelena:
Not merge-ready.

For discussion:
- I recommend updating to docker-compose version 3.  I really prefer using docker swarm for stack deployment.  (the 'docker stack deploy ...' commands deploy as a service to swarm and restart containers if they die).  Let me know if you have a reason not to do this
- while trying to get this running locally and ready for deploy to staging, I pulled in some simple scripts that are similar to what i use on other projects. these are in a new deploy/docker/ directory but i'm happy to move
- i copied your docker-compose to deploy/docker/docker-compose-noaws.yml.  the only real changes were 
  - update docker-compose to 3
  - updated build paths since compose file is in nested dir
  - **for web container the NODE_ENV arg went away** i was getting syntax errors but theres probably a way to put it back in.  That said, i should first understand what i'm trying to do
   - i changed your two `*_REGISTRY_URI` env vars to a single `BT_DOCKER_REGISTRY_URI` env var and suffixed `web` and `worker` in the image names.  This means they need to come from same registry **is that a problem**?

As of now, running ./deploy/docker/deploy-local.sh builds and deploys w/out error but the containers (web, worker, minio/mc) do have errors.  I am not setting any env variables other than what you see in the script itself, so that's the place to start.  I see a bunch of notes in the README related to env vars and production deployment, but i'm not sure which apply to local deployments.  sorry if i missed it!


anyaelena:
@kbenne - related to deployment to internal staging server, i will need to put in a request with cyber for any ports other than 80 and 443 that we want to connect to from off the vm.  It looks like you have mineo running a webserver on port 9000.  Anything else come to mind?

kbenne:
> @kbenne - related to deployment to internal staging server, i will need to put in a request with cyber for any ports other than 80 and 443 that we want to connect to from off the vm. It looks like you have mineo running a webserver on port 9000. Anything else come to mind?

Maybe 27017 for mongo.

deploy/docker/deploy-internalstage.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# NOTE that ic-docker-registry.nrel.gov:5000/ registry server is available only w/in nrel.  also, ssl certs for internal domains are a major hassle.
+# allow your local docker to access ic-docker-registry w/out cert:
+# 1. open docker preferences
+# 2. Daemons: add 'ic-docker-registry:5000' as an insecure registry
+
+#build images
+export BT_DOCKER_REGISTRY_URI=ic-docker-registry.nrel.gov:5000/
+export BT_ENV=stage
+docker-compose -f ./deploy/docker/docker-compose-staging.yml build
```
kbenne: docker-compose-staging doesn't exist yet does it? I do not see it as part of this PR.

deploy/docker/docker-compose-noaws.yml
```diff
@@ -0,0 +1,78 @@
+version: '3'
```
kbenne: With the introduction of this compose file, can we just remove the top level docker-compose.yml file?

kbenne:
@anyaelena I'm not sure if you are aware but there is a .env file that docker-compose respects. located here https://github.com/NREL/alfalfa/blob/develop/.env

deploy/docker/deploy-internalstage.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
kbenne: Just so I'm clear the purpose of this is to push to the staging virtual machine right?

kbenne:
Oh... One more comment. It would be nice to also update the README to reflect these changes.

anyaelena:
@kbenne status update

a few hiccups but the deploy scripts do work now.  Except I *think* that the server node needs to be bound to 0.0.0.0 instead of localhost?  At least, after disk issues i finally have the stack running (for now) on ic-docker-registry.nrel.gov.  the web container isn't responding to that domain, however.  I don't know how your AWS deploys work, but other than those you only deploy on localhost, yes?

Other than this, todo is:
1) possibly update the intended staging vm to have more disc.  or just wait for the new kubernetes setup.
2) minio/mc is just getting constantly restarted.  no harm but we can do better
3) clean up root level docker-compose and .env after making sure all usecases are covered
4) README update

kbenne:
> @kbenne status update
> 
> a few hiccups but the deploy scripts do work now. Except I _think_ that the server node needs to be bound to 0.0.0.0 instead of localhost? At least, after disk issues i finally have the stack running (for now) on ic-docker-registry.nrel.gov. the web container isn't responding to that domain, however. I don't know how your AWS deploys work, but other than those you only deploy on localhost, yes?
> 
> Other than this, todo is:
> 
> 1. possibly update the intended staging vm to have more disc.  or just wait for the new kubernetes setup.
> 2. minio/mc is just getting constantly restarted.  no harm but we can do better
> 3. clean up root level docker-compose and .env after making sure all usecases are covered
> 4. README update

kbenne:
@anyaelena I have only deployed to the local development machine and to aws in the past. There is a script to deploy to aws and the steps are documented here https://github.com/NREL/alfalfa/blob/develop/README.md#if-you-need-to-update-the-worker. It hasn't been exercised in a long time though. Basically you push to the registry and then (re)create the aws task.

PR #43, opened 2020-02-17
change subprocess.call to subprocess.Popen to prevent blocking

corymosiman12:
Only thing I am potentially worried about is that, if somebody were to try and run like a bunch of buildings at the same time, I'm not sure whether the cbi lab host would be able to handle...

Thoughts?

### DIFF
--- a/worker/message_handler.py
+++ b/worker/message_handler.py
@@ -55,9 +55,9 @@ def process_message(message):
                 logger.info('Start simulation for site_ref: %s, and simType: %s' % (siteRef, simType))
 
                 if simType == 'fmu':
-                    subprocess.call(['python', 'stepfmu/step_fmu.py', siteRef, realtime, timescale, startDatetime, endDatetime, externalClock])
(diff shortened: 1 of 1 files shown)


kbenne:
I don't think you want this. If the queue is full and popen does not block, the worker will just keep filling itself up with jobs. Instead the current design is to scale up the number of workers, with each having one job at a time. As this is proposed I don't see anything to stop the working from trying to drain the queue and completely overload itself.

corymosiman12:
@kbenne how would we go about deploying with a version of alfalfa that auto-scales the number of workers?  Is there a specific commit when can look to potentially merge in then with the `cbi` branch to enable this functionality?

kbenne:
@corymosiman12 I have spent a significant amount of time developing a scaling strategy within an AWS environment. After several iterations I settled on having a dedicated "admin" container service that scales up the desired number of workers based on the queue size. 

*however* the cbi lab deployment is not an AWS deployment. I think it is setup to deploy with docker-compose. For the cbi lab I would just run however many workers you have the resources for. All of the workers will just sit there and poll the job queue until there is something to do. You can have as many as you want. `docker-compose scale worker=N`

corymosiman12:
well I didn't know that was a thing...thanks Kyle!

PR #60, opened 2020-04-16
Cbi to replace master 

aviveiros11:
PR the bring master up to date with the latest cbi updates.


corymosiman12:
My only comments:

- as we are not currently using kubernetes, does it make sense to somehow break out the kubernetes deployment stuff into a separate folder / directory...maybe it doesn't matter?

aviveiros11:
> My only comments:
> 
> * as we are not currently using kubernetes, does it make sense to somehow break out the kubernetes deployment stuff into a separate folder / directory...maybe it doesn't matter?
> 
> @anyaelena is this something we want to leave in?

aviveiros11:
@corymosiman12, @anyaelena  I could also maintain the k8's deployment files on a separate branch and remove them from this branch to serve as the new master.  That way we preserve that configuration and can merge in later if need be.

corymosiman12:
That makes sense to me

corymosiman12:
Cool - looks good

PR #66, opened 2020-08-04
Fix running FMUs with Alfalfs

nllong:
The previous code migration seemed to break the FMU running. These changes address running the FMU on the worker node. 

Caveats:
* The start time is always 0 (or the beginning of the FMU year)
* I changed the order of the step_fmu.py arguments to be the same as the step_osm arguments.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,7 +30,7 @@ repos:
     rev: v1.4.4
     hooks:
       - id: autopep8
-        args: ['--in-place', '--aggressive', '--aggressive', '--recursive', '--max-line-length=100', '--ignore=E501,E402,W503,E731']
+        args: ['--in-place', '--aggressive', '--aggressive', '--recursive', '--max-line-length=100', '--ignore=E501,E402,W503,W504,E731']
   - repo: https://github.com/humitos/mirrors-autoflake.git
     rev: v1.3
     hooks:
@@ -40,4 +40,4 @@ repos:
     rev: 3.7.9
     hooks:
       - id: flake8
-        args: ['--ignore=E501,E402,W503,E731,F401']
+        args: ['--ignore=E501,E402,W503,W504,E731,F401']
--- a/alfalfa_worker/worker.py
+++ b/alfalfa_worker/worker.py
@@ -26,6 +26,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import print_function
 
+import traceback
 import json
 import os
 import subprocess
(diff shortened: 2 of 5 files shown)


alfalfa_worker/worker.py
```diff
@@ -86,17 +87,25 @@ def check_step_sim_config(self, message_body):
...
-        start_datetime = message_body.get('startDatetime',
-                                          start)  # TODO change server side message: startDatetime to start_datetime
-        end_datetime = message_body.get('endDatetime',
-                                        end)  # TODO change server side message: endDatetime to end_datetime
+        # TODO change server side message: startDatetime to start_datetime
+        start_datetime = message_body.get('startDatetime', start)
+        if start_datetime == 'undefined':
+            start_datetime = start
+        # TODO change server side message: endDatetime to end_datetime
+        end_datetime = message_body.get('endDatetime', end)
```
nllong, line 95: datetime is weird. FMU wants seconds from 0. I think we should always have datetime as datetime objects, then convert to how OS or FMU wants at time of their execution.

alfalfa_worker/worker.py
```diff
@@ -198,31 +207,62 @@ def step_sim_type(self, p, site_id, step_sim_type, step_sim_value, start_datetim
         :param step_sim_value:
         :param start_datetime:
         :param end_datetime:
+        :param model_type: string, type of model. Choices are 'osm' or 'fmu'
         :return:
         """
-
-        if not os.path.isfile(p):
-            self.worker_logger.logger.info("No file: {}".format(p))
+        self.worker_logger.logger.info("Calling model '{}' step_sim_type '{}'".format(model_type, step_sim_type))
```
nllong, line 213: this code cleanup is to have consistent calls to the subprocess

alfalfa_worker/worker.py
```diff
@@ -290,19 +330,16 @@ def step_sim(self, message_body):
         else:
             site_id = message_body.get('id')
             site_rec = self.ac.mongo_db_recs.find_one({"_id": site_id})
+            self.worker_logger.logger.info('Site record: {}'.format(site_rec))
 
-            # TODO: do we still want default to be 'osm'?
-            sim_type = site_rec.get("rec", {}).get("sim_type", "osm").replace("s:", "")
+            # TODO: Check if simType is defined, error if not.
+            sim_type = site_rec.get("rec", {}).get("simType", "osm").replace("s:", "")
```
nllong, line 336: this was causing simulations to always be OSMs. 😬

.pre-commit-config.yaml
```diff
@@ -30,7 +30,7 @@ repos:
     rev: v1.4.4
     hooks:
       - id: autopep8
-        args: ['--in-place', '--aggressive', '--aggressive', '--recursive', '--max-line-length=100', '--ignore=E501,E402,W503,E731']
+        args: ['--in-place', '--aggressive', '--aggressive', '--recursive', '--max-line-length=100', '--ignore=E501,E402,W503,W504,E731']
```
nllong, line 33: don't require the breaking on long lines at binary operators (or something like that)... this basically allows PyCharm's version of breaking to be allowed in pre-commit.

alfalfa_worker/worker.py
corymosiman12: Nick - it looks like this PR still keeps it as a string.  Do you want to just open this as an issue to track it?

alfalfa_worker/worker.py
corymosiman12: Whoops.  I think I had wanted to get all message passing to utilize snake case, but never implemented on web side of things.  Good catch

alfalfa_worker/worker.py
nllong: yeah, i kept it as a string. Making this an issue would be the best solution for now.

Issue #87, opened 2020-12-18
Add some basic python libraries to the worker


kbenne:
Give the people what they want!

At least....

* symfit
* matplotlib

In addition to what is already available. Perhaps these should be advertised.

corymosiman12:
dude never seen symfit before - looks cool.  Who's using that?  what for?

PR #88, opened 2020-12-18
Add python libs

kbenne:
ref #87


anyaelena:
again, i have tested locally because travis isn't available to us any longer.

PR #93, opened 2020-12-29
Gh actions

### LINKED ISSUE #82: Travis builds are failing due to new new caps on free accounts
`Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting.`

Note that build allowances are shared across the entire NREL org.  We should likely explore alternatives.

anyaelena:
Use GitHub Actions instead of Travis since Travis is no longer reliably free for NREL projects (limited free builds shared across organization).

I also removed the pysindy dependency which is necessary for one user but not broadly applicable.

Fixes #82.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name: Checkout code
-      uses: actions/checkout@v2
+      - uses: actions/checkout@v2
 
       # Runs the Super-Linter action
       - name: Run Super-Linter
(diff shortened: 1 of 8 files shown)


anyaelena:
@kbenne can we just merge this, then add the osw test back in?  As discussed, we are removing pysindy.  If we need in other projects, we'll do it in a branch.

anyaelena:
Merging with osw test commented out.  The test is not passing in develop - we believe it broke related to a python dependency and will address #96 separately.

PR #99, opened 2021-01-07
Update minimum version of rake.

anyaelena:


anyaelena:
This still has intermittent timeouts of integration tests that will often run successfully on restart.  So I'm merging with one instance of passing integration tests.  Develop branch now has timeouts on tests which will help debug what's up with these sometimes hanging tests.

PR #105, opened 2021-01-11
Update mlep

corymosiman12:
- mlep 0.1.1.dev3, which should fix issue with reading a lot of data from bcvtb server: https://github.com/NREL/python-mlep/blob/3d8284d04b551093005c162680a72523911bfeaa/tests/test_mlep_process.py#L33-L49
- break out requirements for py3 and py2
- trim down requirements


alfalfa_worker/Dockerfile
```diff
@@ -48,9 +48,10 @@ RUN curl -SLO $OPENSTUDIO_DOWNLOAD_URL \
     && gdebi -n $OPENSTUDIO_DOWNLOAD_FILENAME \
     && rm -f $OPENSTUDIO_DOWNLOAD_FILENAME
 
-COPY alfalfa_worker/requirements/production.txt /alfalfa/
-RUN  python3 -m pip install --ignore-installed -r /alfalfa/production.txt
-RUN  python -m pip install --ignore-installed -r /alfalfa/production.txt
+COPY alfalfa_worker/requirements/py3-production.txt /alfalfa/
+COPY alfalfa_worker/requirements/py2-production.txt /alfalfa/
+RUN  python3 -m pip install --ignore-installed -r /alfalfa/py3-production.txt
+RUN  python -m pip install --ignore-installed -r /alfalfa/py2-production.txt
```
anyaelena, line 54: @corymosiman12  do you know why we have python2 at all?

anyaelena (approved):
Thank you @corymosiman12 !

alfalfa_worker/Dockerfile
corymosiman12: I believe it is mainly from some of the FMU things, although I'm not 100% positive.  @kbenne ?? Are we able to migrate everything to py3?

Issue #79, opened 2020-12-08
starting simulation with "realtime" toggle hits input validation errors


anyaelena:
````
2020-12-08 16:48:34,516 - alfalfa_worker - INFO - start_datetime type: <class 'str'>	start_datetime: 2020-12-08 09:48:29
INFO:alfalfa_worker:start_datetime type: <class 'str'>	start_datetime: 2020-12-08 09:48:29
2020-12-08 16:48:34,516 - alfalfa_worker - INFO - end_datetime type: <class 'str'>	end_datetime: 2020-12-08 09:48:29
INFO:alfalfa_worker:end_datetime type: <class 'str'>	end_datetime: 2020-12-08 09:48:29
2020-12-08 16:48:34,516 - alfalfa_worker - INFO - realtime type: <class 'bool'>	realtime: True
INFO:alfalfa_worker:realtime type: <class 'bool'>	realtime: True
2020-12-08 16:48:34,517 - alfalfa_worker - INFO - timescale type: <class 'str'>	timescale: 5
INFO:alfalfa_worker:timescale type: <class 'str'>	timescale: 5
2020-12-08 16:48:34,517 - alfalfa_worker - INFO - external_clock type: <class 'bool'>	external_clock: False
INFO:alfalfa_worker:external_clock type: <class 'bool'>	external_clock: False
2020-12-08 16:48:34,517 - alfalfa_worker - INFO - Only one of 'external_clock', 'timescale', or 'realtime' should be specified in message
INFO:alfalfa_worker:Only one of 'external_clock', 'timescale', or 'realtime' should be specified in message
````

PR #118, opened 2021-01-27
Fix broken external clock option

### LINKED ISSUE #117: external clock does not work
Alfalfa just runs a simulation with automatic time advancement

kbenne:
Before this change, the presence of realtime or timescale options in the
job payload would cause the external clock option to be ignored. This
change gives external clock more priority. If external clock is present
then other options are not used.

close #117

### DIFF
--- a/alfalfa_worker/worker.py
+++ b/alfalfa_worker/worker.py
@@ -132,7 +132,10 @@ class Worker:
                 "At least one of 'external_clock', 'timescale', or 'realtime' must be specified")
             sys.exit(1)
 
-        if realtime:
+        if external_clock:
+            step_sim_type = "external_clock"
+            step_sim_value = "true"
+        elif realtime:
             step_sim_type = "realtime"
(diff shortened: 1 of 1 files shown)


anyaelena:
Sorry for the delay @kbenne.  I'm ok to merge this, but ideally we should probably enumerate the combinations of inputs that are allowed.  Why are realtime and/or timescale also set for the test?

kbenne:
I agree it needs a more substantial rework. I'm going to generate some content over in the architecture document about the message payload and go from there. 

realtime and timescale are passed over from the web server where the message packet is first generated. It might be that a better fix is to address it there. Does seem logical to validate the message at the point of creation. I'm not sure that invalidates the change in this PR, but might suggest that the issue doesn't deserve to be closed yet.

anyaelena:
i'll just merge this to unblock your test.  #79 also touches on the broader issue you're describing.

PR #125, opened 2021-02-09
Tests/external clock

corymosiman12:
Add external clocks tests for zipped OSW and OSM

### DIFF
--- a/alfalfa_worker/step_sim/model_advancer.py
+++ b/alfalfa_worker/step_sim/model_advancer.py
@@ -78,12 +78,12 @@ class ModelAdvancer(object):
         self.set_idle_state()
         self.init_sim()
         self.set_db_status_running()
+        self.ac.redis_pubsub.subscribe(self.site_id)
         if self.step_sim_type == 'timescale' or self.step_sim_type == 'realtime':
             self.model_logger.logger.info("Running timescale / realtime")
             self.run_timescale()
         elif self.step_sim_type == 'external_clock':
             self.model_logger.logger.info("Running external_clock")
-            self.ac.redis_pubsub.subscribe(self.site_id)
             self.run_external_clock()
 
     def set_idle_state(self):
--- a/alfalfa_worker/step_sim/model_advancer.py
+++ b/alfalfa_worker/step_sim/model_advancer.py
@@ -79,8 +79,11 @@ class ModelAdvancer(object):
         self.init_sim()
         self.set_db_status_running()
         if self.step_sim_type == 'timescale' or self.step_sim_type == 'realtime':
+            self.model_logger.logger.info("Running timescale / realtime")
             self.run_timescale()
         elif self.step_sim_type == 'external_clock':
+            self.model_logger.logger.info("Running external_clock")
+            self.ac.redis_pubsub.subscribe(self.site_id)
             self.run_external_clock()
 
     def set_idle_state(self):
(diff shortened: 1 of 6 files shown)


alfalfa_worker/step_sim/model_advancer.py
```diff
@@ -78,9 +78,12 @@ def run(self):
         self.set_idle_state()
         self.init_sim()
         self.set_db_status_running()
+        self.ac.redis_pubsub.subscribe(self.site_id)
         if self.step_sim_type == 'timescale' or self.step_sim_type == 'realtime':
+            self.model_logger.logger.info("Running timescale / realtime")
```
anyaelena, line 83: @kbenne @corymosiman12  maybe outside scope for now, but we probably want most logging to be "debug" and to have a configurable option to run alfalfa in debug mode.  At least, i don't think that's configurable now.

alfalfa_worker/step_sim/step_fmu.py
```diff
@@ -151,6 +151,7 @@ def run(self):
             while True:
                 message = self.redis_pubsub.get_message()
                 if message:
+                    print("process_pubsub_message: %s" % message)
```
anyaelena, line 154: Do we really want print instead of logging here?

anyaelena (changes requested):
@corymosiman12  I reject your print statement! (Thank you for the tests.)

tests/integration/test_refrig_case_osw.py
```diff
@@ -42,3 +47,47 @@ def test_simple_internal_clock(self):
...
+    def test_simple_external_clock(self):
+        zip_file_path = create_zip('refrig_case_osw')
+        alfalfa = AlfalfaClient(url='http://localhost')
+        model_id = alfalfa.submit(zip_file_path)
+
+        alfalfa.wait(model_id, "Stopped")
+        start_dt = datetime.datetime(2019, 1, 2, 0, 2, 0)
+        alfalfa.start(
+            model_id,
+            external_clock="false",
```
kbenne, line 60: false? this doesn't look right to me.

alfalfa_worker/step_sim/step_fmu.py
corymosiman12: the `step_fmu` guy doesn't have any logging infrastructure was the only reason I used a print.

alfalfa_worker/step_sim/model_advancer.py
corymosiman12: that makes sense

tests/integration/test_refrig_case_osw.py
corymosiman12: This is a relic of the GraphQL and initially just shipping values to it via the client. I have updated the 'data type' checks on the alfalfa-client side here: https://github.com/NREL/alfalfa-client/blob/7b63d2af7494fc4a1d75ced56d4cb14297876587/alfalfa_client/lib.py#L117, but we are on an older version of the client.

corymosiman12: you can see the old implementation here:

https://github.com/NREL/alfalfa-client/blob/5ecda0c82a8fc3630b60e395b646d59f752c8ca6/alfalfa_client/lib.py#L117

notice because we are using JSON datatypes, the booleans need to be lowercased.

alfalfa_worker/step_sim/step_fmu.py
corymosiman12: removed

alfalfa_worker/step_sim/model_advancer.py
corymosiman12: tracking via https://github.com/NREL/alfalfa/issues/126

corymosiman12:
@anyaelena ready for re-review!

anyaelena (approved):

PR #128, opened 2021-02-09
feat/historian

corymosiman12:
- add influxdb and grafana to stack
- ~~data written ONLY for OSM / OSW (E+) simulations~~ [EDIT] data now available with FMU base simULATIONS
- write data to influxdb as part of `update_db` method
- similar to how we are storing data in mongodb, the site_id is acting as the 'holding container' for all of the point timeseries data. in influxdb, this works by using a _measurement_ (see below for more detail). Said another way, the site_id acts as the table under which all sensor data is written.
- create simple dashboard to display timeseries data
- currently, I don't have it set up so that influx and grafana are optional services - they always get spun up. however, in the `.env`. there is a `USE_HISTORIAN` flag. This is only used by the `OSMModelAdvancer`, which will only write to influx if that value is `true`. 

From [influxdb docs](https://docs.influxdata.com/influxdb/v1.8/concepts/key_concepts/):
_"The measurement acts as a container for tags, fields, and the time column, and the measurement name is the description of the data that are stored in the associated fields. Measurement names are strings, and, for any SQL users out there, a measurement is conceptually similar to a table."_

Screenshot below shows data from TestSmallOfficeOSM::test_simple_external_clock

image


anyaelena (changes requested):
Exciting @corymosiman12 !   A couple thoughts:

1. I wish we could test this - certainly when we hook in haystack history capabilities.  in the meantime can we just query influxdb or something to be sure it's working?
2. Interested in @kbenne's opinion, but I do think we want this to be an optional add to the stack.  So we would have two docker-compose files, one with historian and one without.  We might just add another action in the ci to test each stack...  I can work on this if you prefer.  One of the main selling points of Alfalfa will be the modularity to "pick and choose" which capabilities you want to deploy, so i think this serves as a nice example.  Thoughts?

anyaelena:
Also note that i think it's ok to keep this PR open until after our first release.  I'm really excited that you are pushing this forward, @corymosiman12.

alfalfa_worker/step_sim/model_advancer.py
```diff
@@ -78,9 +81,12 @@ def run(self):
         self.set_idle_state()
         self.init_sim()
         self.set_db_status_running()
+        self.ac.redis_pubsub.subscribe(self.site_id)
```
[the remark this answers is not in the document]
corymosiman12: yes - this is true. at this point, however, fmu_model_advancer isn't being used, we are still using the `RunFMUSite` in the `step_fmu.py` file. I'm assuming part of the effort of migrating everything to Python3 will be to also start porting over to using the base `Model_Advancer` classes for all models, regardless of type.

.env
```diff
@@ -10,3 +10,10 @@ JOB_QUEUE_URL=http://goaws:4100/queue/local-queue1
 MONGO_URL=mongodb://mongo:27017/
 MONGO_DB_NAME=alfalfa
 REGION=us-west-1
+INFLUXDB_DB=alfalfa
+INFLUXDB_HTTP_AUTH_ENABLED=true
+INFLUXDB_ADMIN_USER=admin
+INFLUXDB_ADMIN_PASSWORD=password
+INFLUX_HOST=influxdb
+INFLUX_PORT=8086
+USE_HISTORIAN=true
```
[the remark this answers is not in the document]
corymosiman12: addressed

alfalfa_worker/lib/alfalfa_connections.py
```diff
@@ -35,6 +36,12 @@ def __init__(self):
         self.mongo_db_write_arrays = self.mongo_db.writearrays
         self.mongo_db_sims = self.mongo_db.sims
 
+        # InfluxDB
+        self.influx_db_name = os.environ['INFLUXDB_DB']
```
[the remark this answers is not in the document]
corymosiman12: addressed

corymosiman12:
Hey @anyaelena, check out the newest stuff ( I liked your thoughts ):
- I broke out deployment into a historian specific and normal
- When deploying with the historian, I added wait-for-its for influxdb and grafana to make sure they get up
- I added a duplicate integration test to spin up the historian stack and run against

corymosiman12:
ok well unfortunately there is no good way to specify multiple `--env-file` params with docker-compose ([issue](https://github.com/docker/compose/issues/7847)). so the two deployment options look like:
- `docker-compose up`
- `export HISTORIAN_ENABLE=true && docker-compose -f docker-compose-historian.yml -f docker-compose.yml up`

I couldn't think of a cleaner way to do this...am open to suggestions!

alfalfa_worker/step_sim/osm_model_advancer.py
```diff
@@ -409,9 +414,52 @@ def update_db_with_model_outputs(self):
...
+        Write output data to influx
+        :return:
+        """
+        json_body = []
+        base = {
+            "measurement": self.site_id,
+            "time": f"{self.get_energyplus_datetime()}",
+        }
+        response = False
+        for output_id in self.variables.get_output_ids():
```
kbenne, line 435: @corymosiman12 I don't think you are doing anything that hasn't already been done as we push current values to mongo, but I do wonder how much these get_foo methods are costing us. If ever we need to start performance optimizing this thing, the thing I would be looking at (right after moving current values into redis) would be these get index/id methods and the frequency that they are used. Not suggesting any revision now.

alfalfa_worker/step_sim/step_fmu.py
```diff
@@ -89,11 +83,15 @@ def __init__(self, **kwargs):
         zzip = zipfile.ZipFile(fmupath)
         zzip.extract('resources/kpis.json', self.directory)
 
+        # this shouldn't be hard coded.
+        # step_size in seconds
+        self.step_size = 300
```
kbenne, line 88: yeah. we need to add it as a field of the job message.

kbenne:
@corymosiman12 perhaps it would be a good idea to remove the historian records if the site is removed. We do that for the non historical records here https://github.com/NREL/alfalfa/blob/05d2fc56b03bc6f06ce9d55c36da2aa294e5d4ed/alfalfa_web/server/alfalfa-server.js#L599

On a separate note. I actually think the database schema would should have been designed with the concept of a "model" and a "simulation" instead of site which sort of combines the concepts. This is an idea I'd like to keep in the parking lot.

kbenne (approved):
I think we should merge this, with @anyaelena having discretion if she wants to merge before or after making a first release. I'm starting to think that this feature is so high value, that if we do a release without including this feature, the release might be sidelined for the new shiny thing.

My other comment is that I'm still queasy about introducing another database technology. I've been touting the idea of moving all persistent storage to Postgres, and if that happens, then perhaps TimeseriesDB which is built on top of Postgres would be something to look at https://blog.timescale.com/blog/grafana-time-series-exploration-visualization-postgresql-8c7baa9c3bfe/ 

Awesome work @corymosiman12 !!!

alfalfa_worker/step_sim/osm_model_advancer.py
corymosiman12: yeah, agreed. some of them hit the DB, but most of them actually are just looking at hash maps (this one is just getting hash keys).

alfalfa_worker/step_sim/step_fmu.py
corymosiman12: We could do this the same for the E+ stuff, we would just need to validate a min step size of `60` / only multiples of 60.

corymosiman12:
@kbenne tracking your note about removing data here: https://github.com/NREL/alfalfa/issues/130

anyaelena:
@corymosiman12 nick's went in first and now you have a conflict to resolve.  You all convinced me to include with first release so i'll merge after the conflict is resolved.

corymosiman12:
woohoo! @anyaelena should be good once tests passing

kbenne:
> @corymosiman12 nick's went in first and now you have a conflict to resolve. You all convinced me to include with first release so i'll merge after the conflict is resolved.

peer pressure is a powerful thing!

anyaelena (approved):

PR #132, opened 2021-02-16
Add Poetry and Update Project Dependencies

nllong:
This PR does the following:
* Replace the generic use of `pip` with `poetry` to manage project dependencies. This PR only addresses Python3 related dependencies. Python 2.7 still uses `pip`.
* Update readme regarding use of poetry.
* Update to the latest version of the alfalfa-client for testing. This is presently on a branch and will need to be moved to develop once https://github.com/NREL/alfalfa-client/pull/14 is merged.
* Replace `external_clock` with boolean
* Pass FMU start and end time as datetime string (not seconds). This follows the pattern of the updated alfalfa-client and makes the OSM and FMU interface consistent.


nllong:
Okay, this is ready for review now that the latest version of alfalfa-client has been released.

.github/workflows/ci.yml
```diff
@@ -39,11 +39,12 @@ jobs:
       - name: Install deps
         run: |
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install poetry
```
corymosiman12, line 42: @nllong just fyi if you pip install with poetry: https://python-poetry.org/docs/#alternative-installation-methods-not-recommended. for gh actions, you can also do: https://github.com/NREL/tasty/blob/develop/.github/workflows/ci.yml

corymosiman12:
@nllong we don't have any integration tests on the web frontend to check that what gets sent through the UI correctly posts. Have you tried this manually for OSMs, OSWs, FMUs? I think the datetime stuff and the boolean checks might not work correctly, although I could be mistaken...

.github/workflows/ci.yml
nllong: oh nice. they have a poetry formula! I'll update to that.

nllong:
> @nllong we don't have any integration tests on the web frontend to check that what gets sent through the UI correctly posts. Have you tried this manually for OSMs, OSWs, FMUs? I think the datetime stuff and the boolean checks might not work correctly, although I could be mistaken...

Lemme check

.github/workflows/ci.yml
nllong: i've updated this.

nllong:
> > @nllong we don't have any integration tests on the web frontend to check that what gets sent through the UI correctly posts. Have you tried this manually for OSMs, OSWs, FMUs? I think the datetime stuff and the boolean checks might not work correctly, although I could be mistaken...
> 
> Lemme check

Well, there seems to be some issues with the typed start/end times. I'll have to work on this a bit more apparently.

kbenne:
@nllong If you haven't. Please take a look at the architecture slide deck on teams. There is a slide about the message format and you should update it to reflect the external clock bool change.

poetry looks very pleasant to me!

I feel reluctant about this date time change with fmu. What date time do I give to start at time 0 from the fmu point of view? What time if I want to start fmu 2 minutes into the sim? I didn't fully inspect the step fmu code, but I suspect whatever start date time is given becomes t = 0 and then end time is just elapsed seconds from the start. If that is true, then I can see how it makes the dates more clear from Grafana. But then how would you start an fmu at any time other than 0? The part of this that I do really like is that it eliminates the discrepancy between osm and fmu. The current branches all over the code and treating the start end times as unrestricted strings is a bit gross. The UI certainly needs to be updated to accommodate this change. There is an explicit switch to handle fmu differently on the frontend.

kbenne (changes requested):
@nllong If you haven't. Please take a look at the architecture slide deck on teams. There is a slide about the message format and you should update it to reflect the external clock bool change.

poetry looks very pleasant to me!

I feel reluctant about this date time change with fmu. What date time do I give to start at time 0 from the fmu point of view? What time if I want to start fmu 2 minutes into the sim? I didn't fully inspect the step fmu code, but I suspect whatever start date time is given becomes t = 0 and then end time is just elapsed seconds from the start. If that is true, then I can see how it makes the dates more clear from Grafana. But then how would you start an fmu at any time other than 0? The part of this that I do really like is that it eliminates the discrepancy between osm and fmu. The current branches all over the code and treating the start end times as unrestricted strings is a bit gross. The UI certainly needs to be updated to accommodate this change. There is an explicit switch to handle fmu differently on the frontend.

nllong:
Okay @kbenne @anyaelena @corymosiman12 -- this is ready to 're-review'. This only add poetry and upgrades some dependencies. This uses a new release of the alfalfa-client which does *not* include the bool external_clock changes nor the hardtyped start/end times.

kbenne (approved):

kbenne:
I think there is something good about the date/time change on the fmu, but this is nice and clean to have it in two PRs. Let's see what @corymosiman12 has to say about the fmu time change.

corymosiman12:
🌭  🌭 

@nllong - would you strike through in the original comment at the top of the PR for things that specifically did NOT get implemented, as it's a bit confusing as to what went into this PR now just by reading the GH thread - hopefully just makes our lives easier down the line. Thanks for doing the poetry stuff!

pyproject.toml
```diff
@@ -0,0 +1,36 @@
...
+symfit = "0.5.3"
+# These libraries are used for the data generator/testcase. They are pinned to older versions
+# as the min python version is 3.7
+pandas = "0.24.2"
+scipy = "1.5.4"
+numpy = "1.19.5"
+
+[tool.poetry.dev-dependencies]
+alfalfa-client = "0.1.0.dev7"
+#alfalfa-client = {git = "https://git@github.com/nrel/alfalfa-client.git", rev = "develop"}
```
corymosiman12, line 24: @nllong can you delete this?

nllong: delete the git line? I'd like to keep it that way I don't have to look up the syntax the next time I want to test a new alfalfa-client release.

corymosiman12: got it ok yeah cool schmart schamart

nllong:
> 🌭 🌭
> 
> @nllong - would you strike through in the original comment at the top of the PR for things that specifically did NOT get implemented, as it's a bit confusing as to what went into this PR now just by reading the GH thread - hopefully just makes our lives easier down the line. Thanks for doing the poetry stuff!

done

PR #133, opened 2021-02-17
Publish Docker images

### LINKED ISSUE #114: updated release process
Including Docker image publishing from CI

anyaelena:
- Automate publishing of Docker images to nrel/alfalfa-worker and nrel/alfalfa-web.  
- All commits to develop branch are pushed to Dockerhub with `develop` tag.
- All commits to main are published with a version tag defined [here](https://github.com/NREL/alfalfa/blob/develop/pyproject.toml). (@kbenne @nllong agree that this is where we keep the version?)  `latest` is also updated
- Updates URIs in .env for web and worker images
- Adds VERSION_TAG to .env and replaces hardcoded `latest` in docker-compose.yml

Closes #114


.github/workflows/parse_version.py
```diff
@@ -0,0 +1,5 @@
+import toml
```
nllong, line 1: nice!

.github/workflows/ci.yml
```diff
@@ -136,3 +136,33 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      # Install only toml, which is used to extract version.
+      # We don't do anything else with Python in this job.
+      - name: Install toml
+        run: python -m pip install toml
+
+      - name: Build & Publish Images
+        shell: bash
+        run: ./.github/workflows/publish_to_docker.sh
```
nllong, line 165: this works for now, but we use the GH action script for this in SEED: https://github.com/SEED-platform/seed/blob/develop/.github/workflows/publish.yml#L63-L70. Just an idea for the future.

nllong (approved):
look marvelous!

.github/workflows/ci.yml
anyaelena: Thanks!  We build and push two containers based on a bunch of config in the docker-compose, so i didn't see a straightforward action for this.

anyaelena:
@nllong we decided that the pyproject.toml is specific to worker and will end up moved into that dir in #134 and have version removed in #135.  Instead, we'll add a simple version.txt (or json) in the root.

PR #145, opened 2021-02-23
requirements.txt no longer used for Python 3.

anyaelena:


nllong (approved):
Nice catch, yes, I should have deleted this file!

I thought we were using tox, but if we aren't, then yes remove it. I do highly recommend using tox for CI if we can.

Issue #152, opened 2021-03-03
app.bundle.js size jumped from ~0.5M to ~14.5M in web container


anyaelena:
This is breaking the web server (404) on my K8s-based deployments although it's not clear why.  Docker-compose deployments still work, which means all our tests are still passing. 

I don't know exactly when this changed, but i have an image published from a branch on 12/15/2020 with the smaller file.  

Since we haven't been working on the web container, the only candidate commits are [this dependency update](https://github.com/NREL/alfalfa/commit/4ef15135c5d16c2bfb8e1267176374a1becb42b1) and [this pre-commit commit](https://github.com/NREL/alfalfa/commit/95bbad8e0f31fb5e24f0fddcc626c8b4f6d34a0b)

 In any case, the file should be smaller so we'll start there.

anyaelena:
Ok.  We've been building in development mode, although that's not something that changed recently.  For now, i'm changing this to production.  There's a lot of slimming possible in the web image, but my ambitions are limited this week.

anyaelena:
Building the app with NODE_ENV=production does bring the app.bundle.js down to a size that works on my k8s deployments.  it's still 2M though which is kind of outrageous AND confusing in that the size exploded at some point while we were not actively developing the web container.  I'm closing this since the immediate issue has resolved.  I've opened #156 for a general audit of what's up in the neglected web container.  Note that i did a simple webpack-bundle-analyzer pass and say material-ui as the main bloat... but a cursory look at how we are importing those resources looks consistent with their recommendations. 🤷

PR #153, opened 2021-03-03
Slim(mer) web

anyaelena:


anyaelena:
https://github.com/NREL/alfalfa/pull/153/checks?check_run_id=2027739488#step:7:1968

anyaelena:
This is related to #152 and should help.  It does not explain what went wrong on the published develop image and i am also opening a ticket for a deeper dive into app size in the web container.

PR #154, opened 2021-03-04
Avoid pulling influxdb 2.0.x for now.

anyaelena:


anyaelena:
@corymosiman12  got there first with #155

PR #164, opened 2021-03-05
Support file posts on deployments that aren't local

### LINKED ISSUE #161: webserver index.js needs a url that works for both k8s and docker-compose deployments
this is the relevant commit...

https://github.com/NREL/alfalfa/commit/98a6a3e3352e462f6876d125450cd364e153a5da#diff-175385239b86515c0ba3be4adfadd0bc48b83b4ffea473d1d9f2910c790a7357

anyaelena:
Don't magically guess the url for posting.  Use S3_URL_EXTERNAL if set or default to S3_URL.  As far as i know, it's only when running accessing the stack through localhost that the alternate, externally accessible, url is needed.    I tested this running locally via docker-compose (using the .env file in repo) and on the k8s deployment which accesses minio via a dedicated url (and doesn't set S3_URL_EXTERNAL).

Closes #161 

@nllong @kbenne This is definitely one i want to hear from you on.


nllong (approved):
looks good.

kbenne (approved):
I think this is an improvement. I wonder though if it should just be reduced down even more to a single environment variable `S3_URL`. 

All of this ridiculousness was introduced in an attempt to capture three scenarios.

1. Local development and testing. A developer is working on development, testing uploads through their browser manually, or by executing tests from their development host PC.
2. A real deployment where S3_URL needs to be a real domain name.
3. I originally had in mind that tests would be run through a test container, in which case S3_URL is not localhost, but instead `minio`. Of course if S3_URL is minio then the file server is not addressable from the developer's host PC, so use case 1 is off the table. We do not currently have tests setup to run within a test container so I kind of think this is a non issue. Also if this use case (3) is desired you can always set the URL to minio, and just accept that it will not also work if you try to post from your host PC / browser. (hope this all makes sense)

Bottom line I approve. If any change it might be to just consolidate S3_URL and S3_URL_EXTERNAL because I think this was all just to support a use case that is not being used right now.

anyaelena:
@kbenne I don't agree that usecase 1 is off the table... We want to be able to interact directly with a stack deployed locally via localhost, no???

kbenne:
> @kbenne Thanks for the feedback. I don't agree that usecase 1 is off the table... We want to be able to interact directly with a stack deployed locally via localhost, no???

Oh. I meant 3 is out. or at least not in focus. 1 is very much important to all of us developers! :)

anyaelena:
Ok.  To be explicit, using only S3_URL doesn't work for local deployments.

kbenne:
it would if its value was localhost. right?

anyaelena:
For posterity, i'll summarize our chat: nope, localhost doesn't work for minio internal connections.

Issue #185, opened 2021-09-01
Worker builds fail 


anyaelena:
the Dockerfile command 
```` 
/bin/sh -c python3 -m pip install --upgrade pip setuptools &&      python3 -m pip install poetry &&      python3 -m poetry config virtualenvs.create false &&      python3 -m poetry install --no-dev
````
now fails with 
```` 
The current project's Python requirement (2.7.15+) is not compatible with some of the required packages Python requirement:
    - numpy requires Python >=3.6, so it will not be satisfied for Python 2.7.15
  
````
Poetry requires python = "^3.6.1" so I'm not clear why it's pulling Python 2 and why that changed.  Will follow up soon.

PR #186, opened 2021-09-03
lock version of poetry and force install nodejs

nllong:
* Newer versions of poetry cause this build script to break, locking version for now.
* Need to force install node


anyaelena (approved):

anyaelena:
Fixes #185

PR #219, opened 2022-01-31
Break out files into worker fmu and worker openstudio folders

nllong:
Currently, the contents of the files are the same, but that won't be the case for long.


.env
```diff
@@ -10,6 +10,8 @@ S3_URL_EXTERNAL=http://localhost:9000
 S3_BUCKET=alfalfa
 REDIS_HOST=redis
 JOB_QUEUE_URL=http://goaws:4100/queue/local-queue1
+JOB_QUEUE_URL_1=http://goaws:4100/queue/local-queue1
+JOB_QUEUE_URL_2=http://goaws:4100/queue/local-queue2
```
TShapinsky, line 14: Is there a reason to not make the local queue names more descriptive?

alfalfa_worker/worker_openstudio/add_site.py
```diff
@@ -0,0 +1,296 @@
...
+        with open(points_json_path, 'w') as fp:
+            json.dump(points_json, fp)
+
+        with open(mapping_json_path, 'w') as fp:
+            json.dump(mapping_json, fp)
+
+        # add points to database
+        self.ac.add_site_to_mongo(points_json, self.upload_id)
+
+    def add_osm(self):
```
TShapinsky, line 144: Is this OSM function purely internal? or am I going to get rid of it soon?

.env
nllong: yes, i think we'll want to make the local-queues more descriptive such as local-fmu, local-openstudio...

alfalfa_worker/worker_openstudio/add_site.py
nllong: OSM is external too and yes, we want to get rid of it. It will be a nice code cleanup!

TShapinsky (approved):

PR #223, opened 2022-01-31
rename osmName to modelName to support osm and fmu

nllong:
Simple javascript and python variable rename.

Adding additional queue based on model type next.


nllong:
I need to update alfalfa client to support this change. stay tuned.

PR #225, opened 2022-01-31
move dbops.js out of generic lib folder

### LINKED ISSUE #220: Move dbops to web container only
The `dbops.js` file is only used by the web container and can be moved into that folder. It appears that this file was designed to be extended by clients but I don't think it is needed functionality at the moment.

https://github.com/NREL/alfalfa/blob/develop/lib/dbops.js

nllong:
Fixes #220


kbenne:
I think that as a result of this change you can change the docker-compose context to be the root directory of each container, whereas at the moment the context is the root of the project. https://github.com/NREL/alfalfa/blob/develop/docker-compose.dev.yml#L54, which I think is a bit atypical.

kbenne (approved):

Issue #201, opened 2022-01-19
Remove OSM support from Alfalfa (use OSW in future)


nllong:
Right now there are three paths through alfalfa's worker. OSM, OSW, and FMU. The OSM uses a hardcoded weather file and really doesn't provide utility. This task is to remove the OSM path. Users can still use OSW to run OS/E+ simulations.

PR #227, opened 2022-02-02
Remove .osm workflow from workers

TShapinsky:
#201 
There is still some OSM language existing in alfalfa. Some is dealing with the seed models for OSWs and some is on the front end. I haven't touched the front end due to being unclear as to what is actually referring OSMs and what is an OSW workflow condition that is called "osm".


alfalfa_worker/worker_fmu/add_site.py
```diff
@@ -141,38 +139,6 @@ def insert_os_tags(self, points_json_path, mapping_json_path):
         # add points to database
         self.ac.add_site_to_mongo(points_json, self.upload_id)
 
-    def add_osm(self):
```
nllong, line 144: 🧹!

nllong (commented):
Looking good. Note that #223 will remove some of the osmName items. I haven't merged yet because we'll need to release an updated version of alfalfa-client.

TShapinsky:
@nllong just merged in the osm rename, everything appears to be working on my end

nllong:
> @nllong just merged in the osm rename, everything appears to be working on my end

noice! this is going to be an amazing PR with nice cleanup!

nllong:
@TShapinsky -- is this still a WIP, or can we merge?

TShapinsky:
> @TShapinsky -- is this still a WIP, or can we merge?

I think this is good if we're ready to merge the client change as well

nllong:
sweet, let me do a review of it by tomorrow's meeting. 

I'm happy merging the client down, but will want to wait to release. Do you think it is an issue having people (via poetry) use the Github version of the client?

nllong:
Also, I think [the workflow.tar.gz](https://github.com/NREL/alfalfa/blob/201-remove-osm-workflow/alfalfa_worker/workflow.tar.gz) can be removed since it was used in the OSM workflow? True?

That also means `def extract_workflow_tar(self):` can disappear.

TShapinsky:
> Also, I think [the workflow.tar.gz](https://github.com/NREL/alfalfa/blob/201-remove-osm-workflow/alfalfa_worker/workflow.tar.gz) can be removed since it was used in the OSM workflow? True?
> 
> That also means `def extract_workflow_tar(self):` can disappear.

I can make those changes and see if the tests still pass

TShapinsky:
@nllong we also still have `sim_osm.py` and `step_osm.py` and `osm_model_advancer` should I change all of those to osw?

nllong:
> @nllong we also still have `sim_osm.py` and `step_osm.py` and `osm_model_advancer` should I change all of those to osw?

I would leave those as is because I'm going to rename the files to just "simulate.py" and "step.py" or something like that. That way the worker "API" just knows to call simulate in the shared mounted directory for FMU or OSWs.

TShapinsky:
@nllong added the changes, everything appears to still work

nllong:
> 



> @nllong added the changes, everything appears to still work

sweet, can you fix the pre-commit checks... probably formatting.

TShapinsky:
> > 
> 
> > @nllong added the changes, everything appears to still work
> 
> sweet, can you fix the pre-commit checks? probably formatting.

Looks like I didn't have pre-commit set up on my machine. Should pass now

nllong (approved):
looks great! Thanks!

PR #235, opened 2022-02-17
Update alfalfa-client branch

nllong:
Use the updated develop branch of the alfalfa-client which now has the osmName->modelName committed.


TShapinsky (approved):
Nice to have the client aligned now

Issue #228, opened 2022-02-03
[2/18] Create Dispatch Class for Workers


laurenklun:
toby assisting

Issue #233, opened 2022-02-15
many to one relationship between runs and uploaded models 


anyaelena:
Uses cases:
- multiple simultaneous sims based on single upload
- if model is re-run, each run will need a unique identifier

PR #239, opened 2022-02-18
Create dispatcher class

### LINKED ISSUE #224: Breakup queue on front end
Enable the front end to support multiple worker queues. For example if the modelType (`simType` in JS) is FMU then submit the model to queue1, if OpenStudio then submit to queue2.

nllong:
`__main__.py` previously called the `worker.py` class directly and then `worker.run()`. This PR introduces an additional layer to help dispatch the work through a new class called Dispatcher. `__main__.py` now calls the dispatcher to run.

Items to still address in the form of cleanup
* There are redundant methods in the sim_*, step_*, and add_site.py files. The dispatcher is the only place now where conditional statements pivoting on osm vs fmu should occur.
* rename the "sim_fmu", "sim_osm", etc to just "sim", etc. Remove the job type from the file, that is implied in the directory.

Other items this PR addresses:
* `AlfalfaConnections` is now a base class. When accessing connections you no longer call `self.ac.<xyz>` rather just `self.<xyz>`.
* There is a logger mixin base class that allows classes that need logging to multi-inherit from. Similar to above, no longer do you need to call `self.model_logger.logger`, just `self.logger`.
* Removed the initial work for multiple queues. Only 1 queue now.

Closes #224 with no action
Addresses #228 
Addresses #211


alfalfa_worker/__init__.py
```diff
@@ -12,6 +10,7 @@
 finally:
     del get_distribution, DistributionNotFound
 
+# TODO: remove these, I don't even think they exist as shown anymore!
 # Not sure that these are needed here. Leave for now...
 __all__ = ['add_site',
```
nllong, line 15: I'm pretty sure we can delete these lines with `__all__`

alfalfa_worker/__main__.py
```diff
@@ -23,34 +23,29 @@
...
-else:
-    print(f"Unknown queue in env var of JOB_QUEUE_URL with {os.environ.get('JOB_QUEUE_URL', None)}")
+from alfalfa_worker.dispatcher import Dispatcher
+
 
 if __name__ == '__main__':
     try:
-        worker = Worker()
-        worker.worker_logger.logger.info("Worker initialized")
+        dispatcher = Dispatcher()
```
nllong, line 39: The dispatcher is now in charge of handing the simulation to the appropriate worker runner code. Worker is a misnomer a bit because it isn't a separate node/pod.

alfalfa_worker/dispatcher.py
```diff
@@ -0,0 +1,139 @@
...
+        self.worker_openstudio_class = WorkerOpenStudio()
+        self.worker_fmu_class = WorkerFmu()
+
+    def determine_worker_class(self, test_str):
+        """This is the only place where we should decide
+        if the work is an OSW or an FMU
+        """
+        worker_class = None
+        test_str = test_str.lower().replace('.', '')
+        if test_str in ['osw', 'osm', 'zip']:
```
nllong, line 67: OSM is still needed because after a site is added, the run_sim passes an OSM as the job type, which hits this method.

alfalfa_worker/dispatcher.py
```diff
@@ -0,0 +1,139 @@
...
+                    if action == 'init':
+                        # For testing purposes, just return the
+                        # worker class that was assigned
+                        return worker_class
+                    elif action == 'addSite':
+                        worker_class.add_site(message_body)
+                elif action in ['runSite', 'runSim']:
+                    # get the site ID out of the message
+                    site_id = message_body.get('id')
+                    site_rec = self.mongo_db_recs.find_one({"_id": site_id})
```
nllong, line 108: sorry, had to pull this from mongo. This is also in each of the run* files. This method should get abstracted regardless, perhaps in a "Site" class (similar to an ORM), and then we can easily replace with Postgres or the like.

alfalfa_worker/lib/__init__.py
```diff
@@ -1,3 +1,4 @@
-from .precheck_argus import *  # noqa
-from .make_gzip_file import *  # noqa
-from .tagutils import *  # noqa
+# TODO: these should be imported into the files that need them
```
nllong, line 1: I try not to initialize a bunch of files. I updated the includes in the files that needed them.

alfalfa_worker/lib/alfalfa_connections_base.py
```diff
@@ -12,7 +12,7 @@
 from influxdb import InfluxDBClient
 
 
-class AlfalfaConnections:
+class AlfalfaConnectionsBase(object):
```
nllong, line 15: need to inherit from `object` in python2 for base classes. I look forward to moving this all to python 3.

alfalfa_worker/model_advancer_base.py
```diff
@@ -27,14 +27,17 @@
...
 # sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
-from alfalfa_worker.lib.alfalfa_connections import AlfalfaConnections
-from alfalfa_worker.model_logger import ModelLogger
+from alfalfa_worker.lib.alfalfa_connections_base import AlfalfaConnectionsBase
+from alfalfa_worker.lib.logger_mixins import ModelLoggerMixin
 
 
-class ModelAdvancer(object):
-    """Base class for advancing models"""
+class ModelAdvancerBase(ModelLoggerMixin, AlfalfaConnectionsBase):
```
nllong, line 34: Yes this looks like multiple inheritance and it kinda is. Mixins will call the remaining `__init__` methods.

alfalfa_worker/model_advancer_base.py
```diff
@@ -105,16 +104,16 @@ def run(self):
         self.set_idle_state()
         self.init_sim()
         self.set_db_status_running()
-        self.ac.redis_pubsub.subscribe(self.site_id)
+        self.redis_pubsub.subscribe(self.site_id)
```
nllong, line 107: notice that we no longer have to say `self.ac` or `self.model_logger.logger`. So much nicer!

alfalfa_worker/worker_fmu/add_site.py
```diff
@@ -253,18 +253,18 @@ def add_fmu(self):
...
         """
-        self.add_site_logger.logger.info("add_fmu for {}".format(self.key))
+        self.logger.info("add_fmu for {}".format(self.key))
 
-        self.ac.s3_bucket.download_file(self.key, self.fmu_path)
+        self.s3_bucket.download_file(self.key, self.fmu_path)
 
         # External call to python2 to create FMU tags
-        call(['python', 'lib/fmu_create_tags.py', self.fmu_path, self.file_name, self.fmu_json])
+        call(['python', 'alfalfa_worker/lib/fmu_create_tags.py', self.fmu_path, self.file_name, self.fmu_json])
```
nllong, line 261: I remove the `os.chdir` at the beginning of the `worker.py` which means that we need to path from the root of the python worker project.

alfalfa_worker/worker_fmu/step_fmu.py
```diff
@@ -58,7 +60,7 @@ def __init__(self, **kwargs):
         # Use arbitrary start datetime for now
         self.current_datetime = datetime(1970, 1, 1, 0, 0, 0)
 
-        self.site = self.ac.mongo_db_recs.find_one({"_id": self.site_id})
+        self.site = self.mongo_db_recs.find_one({"_id": self.site_id})
```
nllong, line 63: should be in a related class with "site" data.

tests/worker/test_dispatcher.py
```diff
@@ -0,0 +1,77 @@
...
+            dispatcher.determine_worker_class('oSu')
+        self.assertTrue('Unable to determine worker class of string osu' in str(context.exception))
+
+    def test_worker_add_site(self):
+        dispatcher = Dispatcher()
+
+        # This is a mocked message, if the queue changes, this needs to be updated
+        # to support the new format of the queue messages.
+        # TODO: make the body of the messages at a minimum a dictionary, or create a helper method
+        message_stub = MessageMock({
```
nllong, line 36: New `MessageMock` class can act like the queue for testing.

alfalfa_worker/dispatcher.py
anyaelena: #233 will impact this as well i think?

anyaelena (dismissed):
@nllong  I really appreciate the detailed review.

alfalfa_worker/dispatcher.py
nllong: yes it will. I think we need to make a `Site` class that knows how to talk to the database/mongo. This is starting to look like an ORM/ODM need...

.env
```diff
@@ -10,8 +10,6 @@ S3_URL_EXTERNAL=http://localhost:9000
 S3_BUCKET=alfalfa
 REDIS_HOST=redis
 JOB_QUEUE_URL=http://goaws:4100/queue/local-queue1
```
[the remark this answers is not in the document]
nllong: I just went back to the original name, we could change...

alfalfa_worker/dispatcher.py
```diff
@@ -0,0 +1,137 @@
...
+        """Listen to queue and process messages upon arrival
+        """
+        self.logger.info("Entering dispatcher run")
+        while True:
+            # WaitTimeSeconds triggers long polling that will wait for events to enter queue
+            # Receive Message
+            try:
+                messages = self.sqs_queue.receive_messages(MaxNumberOfMessages=1, WaitTimeSeconds=20)
+                if len(messages) > 0:
+                    message = messages[0]
```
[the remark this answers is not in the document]
nllong: yeah, the SQS queue is FIFO and atomic.

alfalfa_worker/worker_fmu/add_site.py
nllong: I hope that this disappears when we move to python3 and classify all of the classes.

alfalfa_worker/dispatcher.py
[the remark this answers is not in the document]
nllong: yup

PR #243, opened 2022-02-24
Increase timeout for mlep to 1 min

### LINKED ISSUE #242: OSW won't start
Some osw based models will not start simulating. What up with that?

kbenne:
Larger models need more time for EnergyPlus to go through startup/warmup.
1 minute is arbitrary but probably enough for most models.

close #242


anyaelena (approved):
Thanks @kbenne !

PR #277, opened 2022-03-28
Fixed datetime picker

axelstudios:
# Description
Luxon had to be downgraded to v1 to satisfy material-ui-pickers v3 requirements

image


anyaelena (approved):
Thanks @axelstudios !

Issue #276, opened 2022-03-25
start should accept timestep as input


anyaelena:
Timestep is currently hardcoded for both E+ (1 min) and FMUs (5 min).

This will require coordinated update to web, worker, & client

anyaelena:
back to ice box for prioritization

Issue #262, opened 2022-03-17
FMU start time hard coded


anyaelena:
I think [this](https://github.com/NREL/alfalfa/blob/develop/alfalfa_worker/worker_fmu/step_fmu.py#L61) just needs to be set to the startTime. Otherwise the real startTime just gets [overwritten](https://github.com/NREL/alfalfa/blob/develop/alfalfa_worker/worker_fmu/step_fmu.py#L276) at the first time step.

PR #280, opened 2022-03-29
263 fix fmutimescale

### LINKED ISSUE #263: FMUs should respect timescale input
https://github.com/NREL/alfalfa/blob/develop/alfalfa_worker/worker_fmu/step_fmu.py#L171

Note that not only is sleep hardcoded, but we don't account for the variable time for advancing model.  We should rework this step sim logic (probably for osw as well) to truly track real time or maintain the intended timescale without drift.  Probably checking against the clock time when sim starts.

anyaelena:
Fixes #263  and #262. Needs cleanup though!!

@nllong note that this also updates the hardcoded fmu timestep from 300s to 60s.  This brings it into line with hardcoded E+ timestep but pls be aware in case this impacts your work w/CU.  I opened #276 for us to prioritize after the current release.

### DIFF
--- a/alfalfa_worker/worker_fmu/step_fmu.py
+++ b/alfalfa_worker/worker_fmu/step_fmu.py
@@ -192,6 +192,10 @@ class RunFMUSite(AlfalfaConnectionsBase):
                 # update stop flag from db
                 self.db_stop_set()
 
+                # update stop flag from endTime
+                if self.simtime >= self.endTime:
+                    self.stop = True
+
                 if self.stop:
                     break
 
@@ -200,9 +204,6 @@ class RunFMUSite(AlfalfaConnectionsBase):
                     next_step_time = next_step_time + self.realworld_timedelta
                     self.advance = False
 
-                # update stop flag from endTime
-                if self.simtime >= self.endTime:
-                    self.stop = True
         self.cleanup()
 
     # Check the database for a stop signal


anyaelena:
(don't judge me for the sloppy prints, pls - will cleanup before merge)

TShapinsky (dismissed):
This looks fine to me. I think we're all looking forward to finally ripping out hardcoded bits.

alfalfa_worker/worker_fmu/step_fmu.py
```diff
@@ -169,12 +169,12 @@ def run(self):
                         self.set_idle_state()
                         break
         else:
-            # first step - AP not sure this is necessary
-            # if self.simtime == self.startTime:
-            #     print("taking first step at {}",datetime.now())
-            #     self.step()
-            #     print("finished first step at {}",datetime.now())
-            #     next
+            # first step takes extra time
```
TShapinsky, line 172: Is this due to warmup in spawn?

anyaelena: I'm not entirely sure.  it seems to happen after the Spawn warmup completes but it was taking a several minutes when i logged it.

anyaelena:
@TShapinsky yes pls - should be what you reviewed yesterday + a couple fixes.

alfalfa_worker/worker_fmu/step_fmu.py
```diff
@@ -164,13 +169,41 @@ def run(self):
...
+            print("in run with current time & next_step_time: ", current_time, next_step_time)
+            self.advance = False
+            while True:
+                current_time = datetime.now()
+
+                if current_time >= next_step_time:
+                    self.advance = True
+                    # sim is not keeping up with target timescale.
+                    # TODO rethink arbitrary 60s behind
+                    if (current_time > next_step_time + timedelta(seconds=60)):
```
TShapinsky, line 189: In the future we could keep track of step times and stop if step run times are routinely take longer than the target times.

alfalfa_worker/worker_fmu/step_fmu.py
```diff
@@ -164,13 +169,41 @@ def run(self):
...
+                if self.advance:
+                    self.step()
+                    next_step_time = next_step_time + self.realworld_timedelta
+                    self.advance = False
+
+                # update stop flag from db
+                self.db_stop_set()
 
+                # update stop flag from endTime
+                if self.simtime >= self.endTime:
```
TShapinsky, line 205: Should this check and the `db_stop_set()` happen before the `if self.stop: break` block? That way you're guaranteed that you won't have a step run after the simulation is supposed to stop.

alfalfa_worker/worker_fmu/step_fmu.py
```diff
@@ -164,13 +169,41 @@ def run(self):
...
-                if self.db_stop_set():
-                    break
+            # first step takes extra time and needs to happen outside timescale loop
+            if self.simtime == self.startTime:
+                print("taking first step at ", datetime.now())
                 self.step()
-                # TODO: Make this respect time scale provided by user
-                time.sleep(5)
+                print("finished first step at ", datetime.now())
+                next
```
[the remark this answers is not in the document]
anyaelena: no.  good catch.

alfalfa_worker/worker_fmu/step_fmu.py
anyaelena: yes, we can probably do better.  Also note that this logic should be shared between fmu and osws.

alfalfa_worker/worker_fmu/step_fmu.py
anyaelena: Yes again.

anyaelena:
@TShapinsky  I pushed my response to your review.  Thank you for that!

alfalfa_worker/worker_fmu/step_fmu.py
TShapinsky: was any code supposed to be pushed to respond to this?

anyaelena: I just pushed another update.  Since the step() is what updates self.simtime, I don't think this makes a difference functionally but I do think it's cleaner and more readable.

TShapinsky (approved):
Looks good to me

PR #307, opened 2022-06-20
Don't Generate Precompressed Assets

axelstudios:
In production the Express server expects precompiled .gz files


anyaelena (approved):
Thanks @axelstudios. Built/deployed locally and confirmed that UI loads fine for me now.

PR #316, opened 2022-06-30
Alfalfa REST API + Documentation

axelstudios:
This PR implements a RESTful API alongside the existing GraphQL API, combined with automated API doc generation.

- Added REST API (`/api/v2/*`)
- Namespaced haystack API (`/haystack/*`)
- Added automated API documentation using Redoc (`/docs`)
- Fixed bug preventing automated simulation progression through UI
- Fixed bugs where dialogs caused WSOD
- Fixed Mongo request race condition bug that prevented useable results (sitesResolver)
- Improved Mongo performance by filtering queries when applicable (sitesResolver)
- Fixed bug where `Alfalfa` header link was only partially clickable
- Fixed text selection in upload filename
- Fixed unhandled error when upload is canceled
- Fixed broken mc container
- Fixed side-scroll bug
- Improved point dialog with recursive mapping to remove Haystack type prefixes
- Added package.json version (returned by API)
- Added missing copyrights
- Renamed `/sites` route to `/models`
- Production builds have been optimized to minify+mangle and exclude sourcemaps
- Fully supports split development with a developer environment on one host and mongo/redis/minio/goaws on another
- Added `LOGGING` toggle to `.env`
- 0 dependency vulnerabilities

Browse to `/docs` for API documentation

image

## Discovered & Known Issues
- There is currently no concept of multiple "runs" for a "model" because modelId == runId.
- Simulation end-time isn't respected
- Some API endpoints remain to be added/documented
- Mongo collection `runs` has incorrect `job_history` for each, `CreateRun` keeps adding up for each new document
- Runs in mongo never go away even if the run is deleted from the UI (Mongo shows 10+, Haystack returns 1)
  - This causes everything to get progressively slower each time a new file is uploaded, since all mongo sites are queried every second
- Completed simulation download links can be incorrect and should use the S3 external URL
- Several properties of the run (e.g. `timescale`) are not persisted in haystack/mongo so they're not available to the API.
- Completed simulations can't be deleted (button is always disabled)


alfalfa_web/generate-docs.js
```diff
@@ -0,0 +1,69 @@
...
+  apis: ["./server/api-v2.js", "./server/api-haystack.js"],
+  definition: {
+    openapi: "3.1.0",
+    info: {
+      title: "Alfalfa",
+      version: version,
+      description:
+        "Alfalfa transforms Building Energy Models (BEMs) into virtual buildings by providing industry standard building control interfaces for interacting with models as they run",
+      "x-logo": {
+        url: "/assets/alfalfa.svg"
```
TShapinsky, line 20: Looks like this file may not have been committed

TShapinsky:
> Runs in mongo never go away even if the run is deleted from the UI (Mongo shows 10+, Haystack returns 1) 

This is because the delete that is called in the api is done through haystack which then only deletes the site from the `recs` not `runs`

alfalfa_web/server/api-v2.js
```diff
@@ -0,0 +1,408 @@
...
+ */
+router.get("/models", async (req, res) => {
+  const docs = {};
+  const cursor = db.collection("runs").find();
+  for await (const doc of cursor) {
+    if (doc) docs[doc._id] = doc;
+  }
+
+  got
+    .post("http://localhost/haystack/read", {
```
TShapinsky, line 90: My feeling is that the new api should not use the haystack api to query data. It should just go straight to mongo itself.

alfalfa_web/server/resolvers.js
```diff
@@ -100,103 +99,69 @@ function runSiteResolver(args, context) {
...
         "params": {
           "run_id": "${args.siteRef}",
-          "timescale": "${args.timescale}",
+          "timescale": ${args.timescale},
           "start_datetime": "${args.startDatetime}",
           "end_datetime": "${args.endDatetime}",
-          "realtime": "${args.realtime}",
-          "external_clock": "${args.externalClock}"
+          "realtime": ${args.realtime},
+          "external_clock": ${args.externalClock}
```
TShapinsky, line 115: I think this change is causing an JSON parse error on the worker.

```
alfalfa_worker  | INFO:alfalfa_dispatcher:Message Received with payload: {
alfalfa_worker  |         "job": "alfalfa_worker.jobs.openstudio.StepRun",
alfalfa_worker  |         "params": {
alfalfa_worker  |           "run_id": "470c9e36-f8a0-11ec-a714-acde48001122",
alfalfa_worker  |           "timescale": undefined,
alfalfa_worker  |           "start_datetime": "undefined",
alfalfa_worker  |           "end_datetime": "undefined",
alfalfa_worker  |           "realtime": undefined,
alfalfa_worker  |           "external_clock": undefined
alfalfa_worker  |         }
alfalfa_worker  |       }
alfalfa_worker  | 2022-06-30 18:13:09,720 - alfalfa_dispatcher - ERROR - Exception while processing message: Expecting value: line 5 column 24 (char 162) with Traceback (most recent call last):
```

anyaelena:
Note that failing test is a one-off '~ replacement' ci error, but the other integration test passes.

PR #325, opened 2022-07-08
Check in run when error in job is experienced

TShapinsky:
- Added `self.checkin_run()` to error handling code.
- Added test which throws error and then confirms that error exists in `jobs.log` that was committed to the run


anyaelena:
@TShapinsky no reviewer assigned but looks good to me.  Ready to merge?

TShapinsky:
Ready to merge!

anyaelena (approved):

PR #333, opened 2022-08-03
pull in latest FMU-related libraries from boptest

nllong:
This is a simple update to pull in the data_generator.py file from BOPTEST's repo. The Dockerfile and step_run update are small changes not related to the pull from BOPTEST.

### DIFF
--- a/alfalfa_worker/jobs/modelica/step_run.py
+++ b/alfalfa_worker/jobs/modelica/step_run.py
@@ -51,7 +51,6 @@ class StepRun(StepRunBase):
         self.tc = TestCase(**config)
 
         # run the FMU simulation
-        self.kstep = 0  # todo remove if not used
         self.simtime = self.sim_start_time
         self.set_run_time(self.simtime)
 
--- a/alfalfa_worker/lib/data/data_generator.py
+++ b/alfalfa_worker/lib/data/data_generator.py
@@ -3,10 +3,10 @@ Created on Apr 25, 2019
 
 @author: Javier Arroyo
 
-This module contains the Data_Generator class with methods to gather
-data within .csv files for a test case. A test case data-set must
+This module contains the Data_Generator class with methods to generate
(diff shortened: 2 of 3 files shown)


alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -51,7 +51,6 @@ def __init__(self, run_id, realtime, timescale, external_clock, start_datetime="
         self.tc = TestCase(**config)
 
         # run the FMU simulation
-        self.kstep = 0  # todo remove if not used
```
nllong, line 54: says "remove if not used". I don't see it being used :)

alfalfa_worker/lib/data/data_generator.py
```diff
@@ -164,17 +162,15 @@ def generate_weather(self,
...
         with open(model_file) as f:
             newText = f.read().replace(str_old, str_new)
 
         with open(model_file, "w") as f:
             f.write(newText)
 
         # Change to Resources directory
-        currdir = os.getcwd()
+        currdir = os.curdir
         os.chdir(self.weather_dir)
```
TShapinsky, line 173: Is there a way we can do this stuff without changing dir?

nllong: I'm sure there is, but I really don't want to touch this code until we can get adequate testing implemented on this file. I actually plan on removing `data_generator` for our [`fmu_runner.py`](https://github.com/urbanopt/geojson-modelica-translator/blob/develop/geojson_modelica_translator/modelica/lib/runner/fmu_runner.py), eventually.

TShapinsky: Makes sense. I currently have the dispatcher reset the CWD every time it gets a new job (due to MLEP changing directory) so this should be fine

TShapinsky (approved):

Issue #110, opened 2021-01-14
Python dependency management for E+ models, including docs


corymosiman12:
Just document:
- The python EMS stuff will be run with Python 3
- Therefore, feel free to use any packages in your EMS that are in the `py3-requirements.txt`

Additionally, document:
- build custom worker w/additional packages
- run E+ through docker as option for modelers to debug python dependencies

anyaelena:
Develop and document workflow for Eric.  Possible `docker run energyplus...` preliminary test to run e+ models through alfalfa worker locally.

anyaelena:
@TShapinsky bump! It sounds like you have a PR ready for this.

PR #335, opened 2022-08-04
Python dependency management

TShapinsky:
Addresses #110 

When a OpenStudio run is created the job checks for any `requirements.txt` files. If any are found it creates a virtual environment in the run directory and installs packages. The E+ idf is then modified to point to that directory via its python path. This could and should be used by energy modelers to get their python dependencies into alfalfa. The main goal is to make it possible to add dependencies without having to create custom images and restarting the server.

The main downside (that I can see) from this change is that models which take advantage of this will have significantly larger than those without a virtual environment. This is because the entire environment is packaged with the run. One alternative would be to install packages worker wide, however, this would mean that packages would need to be installed on every worker and it would introduce the possibility of weird python dependency conflicts between models.

If this gets merged I will add wiki docs to the openstudio section.

### DIFF
--- a/alfalfa_worker/Dockerfile
+++ b/alfalfa_worker/Dockerfile
@@ -38,6 +38,7 @@ RUN apt-get update \
     python3.7-dev \
     python3.7 \
     python3.7-distutils \
+    python3.7-venv \
     python3-pip \
     libblas-dev \
     && rm -rf /var/lib/apt/lists/*
--- a/alfalfa_worker/jobs/openstudio/create_run.py
+++ b/alfalfa_worker/jobs/openstudio/create_run.py
@@ -33,7 +33,7 @@ class CreateRun(Job):
         # If there are requirements.txt files in the model create a python virtual environment and install packaged there
         requirements = self.run.glob("**/requirements.txt")
         if len(requirements) > 0:
-            check_call(["python", "-m", "venv", str(self.dir / '.venv')])
+            check_call(["python", "-m", "venv", "--system-site-packages", "--symlinks", str(self.dir / '.venv')])
             for requirements_file in requirements:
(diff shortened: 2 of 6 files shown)


TShapinsky:
Previous commit tells the venv to use system site packages when possible and to symlink when it can instead of copying from the system python. Since the entire python stack is built before the image is created I think symlinks should continue to work between workers.

anyaelena:
@TShapinsky With this change to venv/run, we should also remove the data-sci Python deps that are being installed on worker to support E+/Python.

nllong (approved):
this looks great. I second Anya's comment about removing dependencies on the worker that aren't needed by every simulation. I have some tweaks in my mongo schema branch to Dockerfile and poetry as well, but it shouldn't conflict...

nllong:
> 

I think another downside is that each worker will have to undergo the `pip install` on `create_run`, meaning that it will take longer to provision new workers to scale out.

TShapinsky:
> > 
> 
> I think another downside is that each worker will have to undergo the `pip install` on `create_run`, meaning that it will take longer to provision new workers to scale out.

Yeah, though this will only happen if the model is using python plugins with dependencies. Also, after the initial download of a package it is cached by the worker so subsequent installs are much faster. 

In my testing the biggest performance hit is actually the time it takes to tar the much larger run and upload it.

TShapinsky:
@nllong I looked at the dependencies we are currently installing and I think most of them are used in the modelica side. So I don't think there's a lot we can remove at this time.

nllong:
> @nllong I looked at the dependencies we are currently installing and I think most of them are used in the modelica side. So I don't think there's a lot we can remove at this time.

no surprised, to be honest. I think this is good to merge.

PR #338, opened 2022-08-11
Fix model advance bug

### LINKED ISSUE #308: Re-enable clock advance assertion loop for OpenStudio external clock simulation test.

TShapinsky:
fixes #308 

> * EnergyPlus sends a message to the external interface and waits for a response (the contents of the response are not important as long as it is in the right format) before continuing the timestep.
> 
> * This means that when alfalfa starts communicating with e+ there is already a packet there waiting to be read.
> 
> * However, alfalfa writes before it reads meaning that by the time it reads there are actually two packets in the buffer, one from the current timestep and one from the next.
> 
> * At some point (this is the part I'm still not sure about) mlep drops one of the packets (possibly due to a buffer overflow).
> 
> * This results in the packet queue decreasing in size from two packets to one. Which in turn looks like e+ jumped a step when it actually didn't.

This PR adds a read at the beginning of the `advance_to_start_time` function to get mlep and e+ in sync.


tests/integration/test_refrig_case_osw.py
```diff
@@ -95,7 +96,7 @@ def test_simple_external_clock(self):
         sleep(65)
```
nllong, line 96:
```suggestion
        sleep(30)
```

Can we reduce this to make tests run a bit faster?

tests/jobs/openstudio/test_openstudio_workflow.py
```diff
@@ -71,16 +71,15 @@ def test_simple_external_clock(mock_dispatcher: MockDispatcher, model_path: Path
     assert start_dt == run.sim_time
     updated_dt = start_dt
 
-    # TODO fix and uncomment advance external clock loop
-    # for _ in range(2):
+    for _ in range(10):
```
nllong, line 74: nice reenabling!

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -116,28 +116,27 @@ def advance_to_start_time(self):
...
+        Write inputs to simulation and get updated outputs.
+        This will advance the simulation one timestep.
         """
+
+        # This doesn't really do anything. Energyplus does not check the timestep value coming from the external interface
         self.ep.kStep += 1
-        self.exchange_data()
+        # Begin Step
+        inputs = self.read_write_arrays_and_prep_inputs()
+        packet = mlep.mlep_encode_real_data(2, 0, (self.ep.kStep - 1) * self.ep.deltaT, inputs)
```
nllong, line 135: should we document what the `2, 0` arguments are?

nllong (approved):
this looks good! intent makes sense on how you moved the `self.ep` commands. All the tests passed locally for me.

@axelstudios -- I had no issues building this from scratch. I called `docker compose rm -f && docker compose -f docker-compose.yml -f docker-compose.dev.yml build --no-cache`

tests/integration/test_refrig_case_osw.py
TShapinsky: I'm fine with this. The 65 figure has been around for a while so I wasn't sure why that number, haha.

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -116,28 +116,27 @@ def advance_to_start_time(self):
...
-        This does not advance the simulation on its own, instead
-        to advance the simulation increment ep.kstep and then call this function
-        See self.step
-        """
-        # Before step
-        inputs = self.read_write_arrays_and_prep_inputs()
-        self.ep.write(mlep.mlep_encode_real_data(2, 0, (self.ep.kStep - 1) * self.ep.deltaT, inputs))
-        # After step
+    def update_outputs_from_ep(self):
+        """Reads outputs from E+ step"""
```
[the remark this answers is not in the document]
TShapinsky: Close. The biggest update is that we are now reading from the E+ socket before we write for the first time. We weren't before which I think was why we were having advance issues.

alfalfa_worker/jobs/openstudio/step_run.py
TShapinsky: We can. They're weird ExternalInterface constants, but it's worth adding a comment, probably.

alfalfa_worker/jobs/openstudio/step_run.py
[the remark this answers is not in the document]
nllong: gotcha, that makes sense.

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -116,28 +116,32 @@ def advance_to_start_time(self):
...
+        Write inputs to simulation and get updated outputs.
+        This will advance the simulation one timestep.
         """
+
+        # This doesn't really do anything. Energyplus does not check the timestep value coming from the external interface
         self.ep.kStep += 1
-        self.exchange_data()
+        # Begin Step
+        inputs = self.read_write_arrays_and_prep_inputs()
+        # Send packet to E+ via ExternalInterface Socket.
```
nllong, line 135: noice, thanks!

nllong (approved):
🔥

Issue #339, opened 2022-08-11
Add test for timescale run stopping at the right time


TShapinsky:
Right now the stop condition on timescale runs appears to be broken. Create a test for this functionality and try to fix the problem

Issue #76, opened 2020-12-03
add test for simulation stopping based on timestamp.


anyaelena:

anyaelena:
@axelstudios @TShapinsky FYI.  Dropping into 0.4.0 release

PR #340, opened 2022-08-11
Timescale fixes

TShapinsky:
Adds conditions to test if end time has been passed in timescale run. Depends on `model_advance_bug` due to sensitivity to model time. 

It would probably be best to have the end_time stop condition to be in the `step_run_base`. This could be done fairly easily, we just need to move the Modelica `step_run` to use datetimes.

#339 #76


alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -73,6 +73,9 @@ def check_simulation_stop_conditions(self) -> bool:
             return True
         if not self.ep.is_running:
             return True
+        if self.get_energyplus_datetime() >= self.end_datetime:
```
nllong, line 76: I think we should make this call a method similar to `modelica/step_run.py`. That way future model runs will implement something similar.

nllong (changes requested):
looks good. one comment that should probably be resolved. Tests passed locally too.

alfalfa_worker/jobs/step_run_base.py
```diff
@@ -150,6 +150,9 @@ def timescale_step_interval(self):
     def time_per_step(self) -> datetime.timedelta:
         raise NotImplementedError
 
+    def get_sim_time(self) -> datetime.datetime:
```
nllong, line 153: nice

alfalfa_worker/jobs/step_run_base.py
```diff
@@ -139,7 +139,7 @@ def run_timescale(self):
                 next_step_time = next_step_time + self.timescale_step_interval()
                 self.advance()
 
-            if self.check_simulation_stop_conditions():
+            if self.check_simulation_stop_conditions() or self.get_sim_time() >= self.end_datetime:
```
nllong, line 142: even better!

nllong (approved):

PR #341, opened 2022-08-11
Fix job_history recording

TShapinsky:
@nllong this should fix the `run.job_history` issues you've been having

### DIFF
--- a/alfalfa_worker/lib/run.py
+++ b/alfalfa_worker/lib/run.py
@@ -28,11 +28,11 @@ class RunStatus(AutoName):
 class Run:
     # A lot of stuff here is done to store in the db. It is messy. If we are sticking with mongo db or switching to something else it would be prettied.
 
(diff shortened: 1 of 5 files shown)


nllong (approved):
looks great. These simulations typically showed 2 CreateRuns and now they are only one. Thanks for fixing this!

Schermata 2022-08-12 alle 07 42 24

alfalfa_worker/lib/run.py
```diff
@@ -28,11 +28,11 @@ class RunStatus(AutoName):
 class Run:
     # A lot of stuff here is done to store in the db. It is messy. If we are sticking with mongo db or switching to something else it would be prettied.
 
-    def __init__(self, dir: Path = None, model=None, _id=None, job_history=[], sim_type=SimType.OPENSTUDIO, status=RunStatus.CREATED, created=None, modified=None, sim_time=None, error_log=""):
+    def __init__(self, dir: Path = None, model=None, _id=None, job_history=None, sim_type=SimType.OPENSTUDIO, status=RunStatus.CREATED, created=None, modified=None, sim_time=None, error_log=""):
```
nllong, line 31: this seems like the most important change.

PR #342, opened 2022-08-12
Pass realtime, external_clock, and timescale as string

nllong:
There have been some exceptions raised saving the messages with types other than strings. This enforces the data to pass as strings, then typed correctly on the other side.

### DIFF
--- a/alfalfa_web/server/resolvers.js
+++ b/alfalfa_web/server/resolvers.js
@@ -83,11 +83,11 @@ function runSiteResolver(args, context) {
         "job": "${job}",
         "params": {
           "run_id": "${args.siteRef}",
-          "timescale": ${args.timescale},
+          "timescale": "${args.timescale}",
           "start_datetime": "${args.startDatetime}",
           "end_datetime": "${args.endDatetime}",
-          "realtime": ${args.realtime},
-          "external_clock": ${args.externalClock}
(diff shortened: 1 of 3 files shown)


alfalfa_web/server/resolvers.js
```diff
@@ -83,11 +83,11 @@ function runSiteResolver(args, context) {
...
           "run_id": "${args.siteRef}",
-          "timescale": ${args.timescale},
+          "timescale": "${args.timescale}",
           "start_datetime": "${args.startDatetime}",
           "end_datetime": "${args.endDatetime}",
-          "realtime": ${args.realtime},
-          "external_clock": ${args.externalClock}
+          "realtime": "${args.realtime}",
+          "external_clock": "${args.externalClock}"
         }
```
TShapinsky, line 91: Do we want the same `undefined` protections in this as in the `api-v2.js`?

TShapinsky (approved):
Looks good to me

alfalfa_web/server/resolvers.js
nllong: oh, with the `!!` and `|| 5`. If so, then yeah, we should be consistent.

TShapinsky (approved):
looks good to merge once tests pass

PR #343, opened 2022-08-12
Add test for automatic virtual environment creation

TShapinsky:
Adds basic test for automatic creation of per model virtual environments


TShapinsky:
It appears that I should not have named my python energyplus plugin `test_plugin.py` because that makes pytest run it as a unit test...

tests/integration/test_small_office_osw.py
```diff
@@ -0,0 +1,30 @@
+import datetime
+
+import pytest
+from alfalfa_client.alfalfa_client import AlfalfaClient
+
+from tests.integration.conftest import create_zip
+
+
+@pytest.mark.integration
```
nllong, line 9: nice!

alfalfa_worker/jobs/openstudio/lib/workflow/measures/python/measure.rb
```diff
@@ -71,7 +71,11 @@ def run(ws, runner, user_args)
 
     # modify python plugin search paths
     ws.getObjectsByType('PythonPlugin_SearchPaths'.to_IddObjectType).each do |o|
-      python_paths = ["../.venv/lib/python3.7/site-packages", "/usr/local/lib/python3.7/dist-packages"]
+      python_paths = ["/usr/local/lib/python3.7/dist-packages"]
```
nllong, line 74: looks good. nice work getting this to work because it looks painful to figure out. Maybe a bit more documentation?

nllong (approved):
looks good. Tests passed!

alfalfa_worker/jobs/openstudio/lib/workflow/measures/python/measure.rb
nllong: And the custom build of EP will require this path to be updated.

nllong (approved):
Fandidiliastic!

PR #352, opened 2022-08-24
Add Mongo ORM/ODM

nllong:
This PR replaced the direction mongodb connections from Python with an object relationship/document mapper (called mongoengine). The main advantage is the validation capabilities added to all data going in and coming out of the database due to a typed scheme

To test this PR, rebuild your test containers with:

```
docker compose rm -f && docker compose -f docker-compose.yml -f docker-compose.dev.yml up --build
```

Then run the unit tests and integration tests in a separate terminal

```
poetry run pytest
poetry run pytest -m "integration"
```

In summary, this PR:
* Add mongo engine ORM
* Creates a database schema per the mongo engine object specification
* Creates a new `site` object which is extracted during the `recs` writing. 
* Creates relationships between objects
* Enables cascading deletes (if called from the python side of the code). `e.g., site.delete() now removes all models, runs, recs, writearrays, etc`
* Moved `add_site_to_mongo` call to the base alfalfa connection class
* Centralized the `stop` method where the mongo cleanup occurs (no longer in each job)
* Adds unit test with basic mock data for mongoengine/mongo


nllong:
Hi @TShapinsky  -- I am stuck at why the Run is not progressing to the next state after CreateRun. I am 95% sure it is in the new code that I had to write to mimic the previous direct MongoDB calls, but I can't find it. 

I am able to consistently recreate by calling:

`poetry run pytest -m "integration" tests/integration/test_refrig_case_osw.py::TestRefrigCaseOSW::test_basic_io`

I think the issue is related to the job/run status being reset to "READY", maybe?

image

TShapinsky:
> Hi @TShapinsky -- I am stuck at why the Run is not progressing to the next state after CreateRun. I am 95% sure it is in the new code that I had to write to mimic the previous direct MongoDB calls, but I can't find it.
> 
> I am able to consistently recreate by calling:
> 
> `poetry run pytest -m "integration" tests/integration/test_refrig_case_osw.py::TestRefrigCaseOSW::test_basic_io`
> 
> I think the issue is related to the job/run status being reset to "READY", maybe?
> 
> image

Thanks @nllong I'll look into this today

TShapinsky:
@nllong 
Here is what I believe the problem to be:

1. [Sites.js](https://github.com/NREL/alfalfa/blob/add-mongoengine/alfalfa_web/components/Sites/Sites.js#L175) gets status information from the site object not the run object.
2. My hack had been to inject run status into the site `simStatus` field using the site [resolver](https://github.com/NREL/alfalfa/blob/add-mongoengine/alfalfa_web/server/resolvers.js#L195)
3. When you changed the relationship between runs and sites it broke that substitution so now the status is always `stopped`
4. Because the client waits for the status to be `READY` before advancing it will never advance past the `CreateRun`

@axelstudios we could fix this by either fixing the bodge or making the page actually pull date from the run objects

TShapinsky:
> @nllong Here is what I believe the problem to be:
> 
>     1. [Sites.js](https://github.com/NREL/alfalfa/blob/add-mongoengine/alfalfa_web/components/Sites/Sites.js#L175) gets status information from the site object not the run object.
> 
>     2. My hack had been to inject run status into the site `simStatus` field using the site [resolver](https://github.com/NREL/alfalfa/blob/add-mongoengine/alfalfa_web/server/resolvers.js#L195)
> 
>     3. When you changed the relationship between runs and sites it broke that substitution so now the status is always `stopped`
> 
>     4. Because the client waits for the status to be `READY` before advancing it will never advance past the `CreateRun`
> 
> 
> @axelstudios we could fix this by either fixing the bodge or making the page actually pull date from the run objects

Well, this could have been an issue. Further digging yields that for some reason no run status is being communicated to the client at all...

alfalfa_web/server/dbops.js
```diff
@@ -93,9 +94,13 @@ function writePoint(id, siteRef, level, val, who, dur, db) {
             });
         } else {
           // In this case the point has never been written to and there is no
-          // existing write array in the db so we create a new one
+          // existing writearray in the db, so error out. A default writearray
+          // should have been created when the rec was created on the backend.
+          console.log("No writearrary found for point " + id + ". Make sure it exists in the original tag file.");
```
nllong, line 99: `writearrays` are now to be propopulated when we parse the tags. This will allow for the frontend to only read/set values on existing records.

alfalfa_worker/dispatcher.py
```diff
@@ -97,17 +97,17 @@ def find_class(path):
...
-        return klazz
+        class_ = getattr(module, components[-1])
+        return class_
 
     @staticmethod
     def print_job(job_name):
-        klazz = Dispatcher.find_class(job_name)
-        print(f"Name: \t{klazz.__name__}")
-        print(f"Description: \t{klazz.__doc__}")
+        class_ = Dispatcher.find_class(job_name)
```
nllong, line 105: just following the python convention.

alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -110,22 +105,27 @@ def get_sim_time(self) -> datetime:
...
+        # look in the database for current writearrays which has an array of controller
         # input values, the first element in the array with a value
         # is what should be applied to the simulation according to Project Haystack
-        # convention
-        for array in self.mongo_db_write_arrays.find({"siteRef": self.run.id}):
-            _id = array.get('_id')
-            for val in array.get('val'):
+        # convention. If there is no value in the array, then it will not be passed to the
+        # simulation.
+        for array in WriteArray.objects(siteRef=self.site.ref_id):
```
nllong, line 114: i really wanted to abstract this out, but the way that modelica handles this is different than openstudio.

alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -196,20 +198,3 @@ def advance(self):
     @message
     def stop(self):
         super().stop()
```
nllong, line 200: this has been moved to the parent class.

alfalfa_worker/jobs/step_run_base.py
```diff
@@ -9,11 +10,25 @@
...
 
 
 class StepRunBase(Job):
-    def __init__(self, run_id, realtime, timescale, external_clock, start_datetime, end_datetime) -> None:
+    def __init__(self, run_id: str, realtime: bool, timescale: int, external_clock: bool, start_datetime: str, end_datetime: str, **kwargs) -> None:
+        """Base class for all jobs to step a run. The init handles the basic configuration needed
+        for the derived classes.
+
+        Args:
+            run_id (str): Run object ID.
```
nllong, line 23: i wanted to document more... but this is just a start.

alfalfa_worker/jobs/step_run_base.py
```diff
@@ -186,6 +207,39 @@ def stop(self) -> None:
         super().stop()
         self.set_run_status(RunStatus.STOPPING)
 
+        # Clear current values from the database when the simulation is no longer running
+        if not self.skip_stop_db_writes:
```
nllong, line 211: here in order to enable certain unit tests that don't need the db. not the prettiest :(

alfalfa_worker/lib/models.py
```diff
@@ -0,0 +1,262 @@
...
+    ListField,
+    ReferenceField,
+    StringField,
+    UUIDField
+)
+
+from alfalfa_worker.lib.sim_type import SimType
+
+
+class TimestampedDocument(Document):
```
nllong, line 23: base class which we inherit all models. this ensures that we have `created` and `modified` fields on all documents.

alfalfa_worker/lib/models.py
```diff
@@ -0,0 +1,262 @@
...
+        return Rec.objects.filter(site=self, **kwargs)
+
+
+class RecInstance(EmbeddedDocument):
+    """This is a flat haystack representation of a point. Right now this stores the union of any
+    value that is expected. Many of these fields are probably not used anymore on this object
+    (e.g., simStatus, step), but they are still included because the code base is reading/writing
+    the valued. For example, it is unsetting `step` even though it doesn't appear to be written to."""
+
+    # TODO: maybe we need to make this a dynamic field: https://docs.mongoengine.org/guide/defining-documents.html#dynamic-document-schemas
```
nllong, line 93: this is an important note. If there is a haystack field that isn't here, then alfalfa won't allow it to persist.

alfalfa_worker/lib/models.py
```diff
@@ -0,0 +1,262 @@
...
+    sp = StringField()
+    hvac = StringField()
+    ahu = StringField()
+    equip = StringField()
+
+    def __init__(self, *args, **kwargs):
+        # if an arg is coming in with `return` then replace it with `return_`
+        # to allow it to persist correctly in the database
+        if 'return' in kwargs:
+            kwargs['return_'] = kwargs.pop('return')
```
nllong, line 152: `return` is a reserved word. so force it to be `return_`

alfalfa_worker/lib/run_manager.py
```diff
@@ -123,83 +126,196 @@ def checkout_run(self, run_id: str) -> Run:
...
+            rec = Rec(ref_id=entity['id'].replace('r:', ''), site=site).save()
+            rec_instance = RecInstance(**entity)
+            rec.rec = rec_instance
+            rec.save()
+
+            # Create default writearray objects for the site. Create this on the backend
+            # to ensure that the links are created correctly and accessible from the frontend.
+            #   Only check for records tagged with writable.
+            #   This may need to be expanded to other types in the future.
+            if rec.rec.writable == 'm:':
```
nllong, line 314: this is where the writearrays are populated in mongo. No longer being created from the frontend. This will likely change to being just Redis???

deploy/k8s-historian.yml
```diff
@@ -403,7 +403,7 @@ items:
             io.kompose.service: mongo
         spec:
           containers:
-            - image: mongo
+            - image: mongo:5.0.12
```
nllong, line 406: forcing version 5, okay?

pyproject.toml
```diff
@@ -41,10 +43,10 @@ influxdb = "^5.3.1"
 # required for tests to pass right now. There are often issues when pinning to a branch
 # and calling poetry update. I have had to completely remove my venv and reinstall in order
 # to get the latest updates.
-alfalfa-client = {git = "https://git@github.com/nrel/alfalfa-client.git", rev = "bc2ba85"}
-# alfalfa-client = { path = "../alfalfa-client", develop = false}
+alfalfa-client = {git = "https://git@github.com/nrel/alfalfa-client.git", branch = "update-response-text"}
```
nllong, line 46: need to change this to a release version once we release the client.

tests/integration/test_simple_thermostat_fmu.py
```diff
@@ -56,7 +56,7 @@ def test_io_with_external_clock(self):
 
         # Attempt to override the measured temp (ie zone temperature),
         # and the setpoint, such that zone temperature is over setpoint.
-        self.alfalfa.setInputs(self.model_id, {"oveWriMeasuredTemp_u": 303.15, "TsetPoint_u": 294.15})
+        self.alfalfa.setInputs(self.model_id, {"oveWriMeasuredTemp_u": 303.15, "oveWriSetPoint_u": 294.15})
```
nllong, line 59: the previous values weren't actually defined in the FMU, but it didn't matter because they weren't needed to trip the control signal for testing.

nllong:
@TShapinsky  -- do you have a bit of time to help figure out why the integration tests are no longer passing! Running the `test_basic_io` is now failing on the "stopSite" command. The issue came about when I tried to get the `pkg_resources` issue fixed on GitHub Actions and now I can't seem to get anything working locally or remotely now. :(. This was the commit that fixed the `pkg_resources` but appeared to break the unit tests (locally) -- https://github.com/NREL/alfalfa/pull/352/commits/f72628e4027f02c430febc2739608384a888fd70

nllong:
Tests should be passing now! Sorry for thinking that this was a deeper problem! 

@axelstudios -- I think we can merge yours, then I can rebase mine on top of yours. Let me know what you think.

axelstudios:
@nllong Sounds good, let's merge mine, rebase, and fix conflicts

alfalfa_worker/lib/models.py
TShapinsky: I wonder if instead of trying to capture an entire all of the possible tags we instead have a `Tag` model which each just have a key, value, and a reference to a `Rec`. I feel like this could also potentially make the transition to postgres easier since it's a common relational database paradigm.

alfalfa_worker/lib/run_manager.py
```diff
@@ -127,95 +135,208 @@ def checkout_run(self, run_id: str) -> Run:
...
+                site.geo_city = entity.get('geoCity')
+                site.geo_state = entity.get('geoState')
+                site.geo_country = entity.get('geoCountry')
+                site.geo_coord = entity.get('geoCoord')
+                site.sim_status = entity.get('simStatus')
+                site.sim_type = entity.get('simType')
+                site.save()
+
+            rec = Rec(ref_id=id, site=site).save()
+            rec_instance = RecInstance(**entity)
```
TShapinsky, line 337: If we made `Tag` a db model we could enumerate through entities here and add them.

alfalfa_web/server/alfalfa-server.js
```diff
@@ -166,23 +166,30 @@ class AlfalfaServer extends HServer {
...
-        const val = r.readScalar();
-        db.add(key, val);
-      } catch (err) {
-        console.log(err);
+    if (rec) {
+      const keys = Object.keys(rec);
+
+      for (let j = 0; j < keys.length; ++j) {
+        const key = keys[j];
+        const r = new HJsonReader(rec[key]);
```
axelstudios, line 183: These lines can be shortened to:
```js
      for (const [key, recValue] of Object.entries(rec)) {
        const r = new HJsonReader(recValue);
```

alfalfa_worker/lib/run_manager.py
nllong: I like this idea, a lot. I just kept the sub-document approach for parity-sake.

alfalfa_web/server/alfalfa-server.js
nllong: updated

TShapinsky: Wow, lookie here at mr code golf

axelstudios (approved):

nllong:
Should we "squash and merge" or just "merge"?

TShapinsky (dismissed):
LGTM

TShapinsky:
I think the `curStatus` in the mock mongo data is causing these errors https://github.com/NREL/alfalfa/blob/add-mongoengine/tests/worker/helpers/mock_mongo_data.py#L25

.github/workflows/ci.yml
```diff
@@ -34,19 +34,34 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
 
+      - name: Start MongoDB
```
TShapinsky, line 37: I'm wondering if the tests which require running mongo should be moved out of regular unit tests. Since so far our unit tests have no requirements for external running software.

tests/worker/test_models.py
```diff
@@ -0,0 +1,106 @@
...
+        assert len(Rec.objects(ref_id__in=['site_456_rec_1', 'site_456_rec_2'])) == 2
+
+        # verify the relationship between the site and the rec
+        site = Site.objects()[1]
+        assert site.ref_id == '456'
+        assert site.recs().count() == 2
+        rec_1 = site.recs()[0]
+        assert rec_1.ref_id == 'site_456_rec_1'
+
+    def test_site_rec_querying(self):
```
TShapinsky, line 92: This will always error now that `curStatus` is no longer in the db model, right?

nllong: I just fixed that... Can you review again and approve if all good?

axelstudios (approved):

TShapinsky (approved):
Looks good, let's merge

nllong:
🥳

PR #353, opened 2022-09-02
Add automatic handling of eplus error logs and test case

TShapinsky:
Push energyplus error log to run `error_log` if energyplus fails to start


nllong (approved):
looks good!

PR #356, opened 2022-09-23
Write to Redis

axelstudios:
This PR moves read/write operations of write arrays, `curVal`, and `curStatus` from Mongo to Redis.

## Worker Updates

- Fixed clearing of `writeErr` when updating rec collection
- Better resolver error handling if an id doesn't exist
- Fixed recs `writeVal` being written as a Haystack string type instead of a number
- Fixed writearrays bug that would return the previous array value rather than the updated value

## Web Updates

- Fixed React routing to redirect to home on unknown routes
- Restricted uploads to `.zip` and `.fmu` file extensions
- Added favicon
- Resolved all high-security vulnerabilities

## What's Changed

- Replaced all usage of Mongo `writearrays` collection to instead use Redis
  - Removed unused `who` field for optimized Redis storage
  - The redis keys for write arrays are namespaced by siteRef and pointId: `site:${siteRef}:point:${pointId}`, and are stored as 17-length redis lists of strings, with empty strings representing null values.
- Replaced all usage of `curVal` and `curStatus` in Mongo's `recs` collection to instead use Redis
  - The redis keys for these properties are namespaced by siteRef and recId: `site:${siteRef}:rec:${recId}`, and are stored as key-value hashes with Haystack string values:
    ```
    {
      "curStatus": "s:ok",
      "curVal": "n:123.45"
    }
    ```
- Changed writable-point default level to 17 to match Project Haystack
  - Fixed bug that didn't restrict level value and could cause huge data overflows


alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -110,18 +110,21 @@ def get_sim_time(self) -> datetime:
...
+        for key in self.redis.scan_iter(prefix + '*'):
+            key = key.decode('UTF-8')
+            _id = key[len(prefix):]
+            write_array = self.redis.lrange(key, 0, -1)
+            for value in write_array:
+                if len(value) > 0:
                     dis = self.id_and_dis.get(_id)
                     if dis:
-                        u[dis] = val
+                        u[dis] = int(value.decode('UTF-8'))
```
TShapinsky, line 127: @axelstudios This appears to be where the test is failing

alfalfa_web/components/App/App.js
```diff
@@ -61,8 +61,9 @@ class App extends React.Component {
             </AppBar>
             <Routes>
               <Route path="/" element={<Upload />} />
-              <Route path="models" element={<Sites />} />
-              <Route path="sims" element={<Sims />} />
+              <Route path="/models" element={<Sites />} />
+              <Route path="/sims" element={<Sims />} />
```
TShapinsky, line 65: Out of scope for this PR, but it would be nice if we could align the terminology/paths with the vocab we use in other parts of alfalfa.

alfalfa_web/server/dbops.js
```diff
@@ -28,114 +20,105 @@ function currentWinningValue(array) {
...
-
 // Get a point by siteRef and display name
 function getPoint(siteRef, name, db) {
   const mrecs = db.collection("recs");
   return mrecs.findOne({ site_ref: siteRef, "rec.dis": `s:${name}` });
 }
 
-function writePoint(id, siteRef, level, val, who, dur, db) {
+async function getWriteArray(siteRef, id, redis) {
+  const key = `site:${siteRef}:point:${id}`;
```
TShapinsky, line 30: Do we want to move this to a function? It gets used enough that if we ever wanted to change this it would be easier to do it all in the same place.

alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -110,18 +110,21 @@ def get_sim_time(self) -> datetime:
...
+        for key in self.redis.scan_iter(prefix + '*'):
+            key = key.decode('UTF-8')
+            _id = key[len(prefix):]
+            write_array = self.redis.lrange(key, 0, -1)
+            for value in write_array:
+                if len(value) > 0:
                     dis = self.id_and_dis.get(_id)
                     if dis:
-                        u[dis] = val
+                        u[dis] = float(value.decode('UTF-8'))
```
TShapinsky, line 127: The nuts and bolts of getting points out of redis should probably happen in the `Job` superclass so that it need not be re-implemented on every job.

alfalfa_worker/lib/alfalfa_connections_base.py
```diff
@@ -47,7 +46,7 @@ def __init__(self):
             self.influx_db_name = None
             self.influx_client = None
 
-    def add_site_to_mongo(self, haystack_json, site_ref):
+    def add_site_to_db(self, haystack_json, site_ref):
```
TShapinsky, line 49: Is this being called from anywhere?

TShapinsky (changes requested):

alfalfa_worker/lib/alfalfa_connections_base.py
axelstudios: It's called in two places:
- [Modelica](https://github.com/NREL/alfalfa/blob/2057c8829a869a89bda6c245ecbddb3e69576f4a/alfalfa_worker/jobs/modelica/create_run.py#L70)
- [OpenStudio](https://github.com/NREL/alfalfa/blob/2057c8829a869a89bda6c245ecbddb3e69576f4a/alfalfa_worker/jobs/openstudio/create_run.py#L131)

TShapinsky: But that calls [this function](https://github.com/NREL/alfalfa/blob/2057c8829a869a89bda6c245ecbddb3e69576f4a/alfalfa_worker/lib/run_manager.py#L182) not the one in alfalfa_connections_base

TShapinsky:
I'm not sure where that last commit came from. I was pulling from this branch but never pushed...

axelstudios:
> I'm not sure where that last commit came from. I was pulling from this branch but never pushed...

That merge showed up because you were on an old commit that I amended: 
image

I'll rebase the history

TShapinsky (dismissed):
If this passes I think it's good to merge

TShapinsky (approved):

PR #359, opened 2022-09-29
Add all haystack tags to the mongo `rec` model

nllong:


TShapinsky (approved):
We can talk about it in the meeting soon. But for this paradigm this does seem like the logical conclusion. The only issue I can think of is the prevalence of custom haystack tags. Right now we generate all the tags ourselves, so that shouldn't be a problem.

PR #364, opened 2022-10-07
Add fix for incorrect timesteps when days roll over

TShapinsky:
fixes #363

PR #365, opened 2022-10-07
Small Optimization for Point Writing

TShapinsky:
We were still writing to mongo for every output point for every step. This seems to provide a speed up in my limited testing of greater than 20 times. I can now run the refrig_test model at over 300x realtime on my machine when I couldn't run at more than 15x before.

I discovered this while working on PR #364

### DIFF
--- a/alfalfa_worker/jobs/openstudio/step_run.py
+++ b/alfalfa_worker/jobs/openstudio/step_run.py
@@ -307,10 +307,6 @@ class StepRun(StepRunBase):
                     rec__cur="m:"
                 )
 
-                # Write to points
-                self.run.get_point_by_key(output_id).val = output_value
-                self.run_manager.update_db(self.run)
-
     def update_sim_time_in_mongo(self):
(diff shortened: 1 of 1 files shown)


axelstudios:
Are there no consequences to skipping this?  That `update_db` call does a number of things, especially updating point values in mongo.

If it's definitely safe to get rid of then we can remove the `get_point_by_key` function, too, since it's no longer used

TShapinsky:
> Are there no consequences to skipping this? That `update_db` call does a number of things, especially updating point values in mongo.

Since we're relaying point values through redis we shouldn't need to update point values in mongo, right? `update_db` is called automatically when other things happen to the run (status update, time change, etc).
> 
> If it's definitely safe to get rid of then we can remove the `get_point_by_key` function, too, since it's no longer used

To me, the ideal function would be that the job doesn't touch redis at all and instead just touches the point object (which it gets via `get_point_by_key`) which then triggers a write or read with redis.

axelstudios (approved):

PR #368, opened 2022-10-13
Switch from zipfile to shutil to improve structure of uploaded models

TShapinsky:
In my testing integration test model zip files carry over the absolute paths from the machine they were created on. This means my zip file has a `Users/Toby/Documents/Projects/alfalfa/tests/integration/models/...` structure which is annoying. This PR switches to shutil for making the archive which is simpler and seems to fix that problem. The issue could probably have been fixed with zipfile, but shutil is probably a better fit anyways.


nllong (commented):
I am assuming that `shutil` is still cross platform, right?

TShapinsky:
> I am assuming that `shutil` is still cross platform, right?

Yes shutil uses the same underlying library as zip for zip file generation (`zlib`)

PR #371, opened 2022-10-20
Demo Docker Layer Caching in CI

TShapinsky:
@axelstudios was talking about this a while ago. The caching action can add a good amount of time to the workflow, however, when there are few modifications to the dockerfile it can definitely provide a bit of a speed up.


TShapinsky:
This seems to (best case) make tests run about 8 minutes faster. In cases where changes have been made to the dockerfiles expect tests to be about 10 minutes slower for the first run.

PR #373, opened 2022-10-31
Change python measure name

TShapinsky:
the measure for adding python environment to the idf was called `python` this is a pretty common measure name for projects using python. If there are two measures with the same directory name things break very badly. So I'm changing the measure name to `alfalfa_python_environment`.


bonnema:
oops! sorry 😀

anyaelena (approved):

PR #378, opened 2022-11-02
move dependencies out of worker Dockerfile

TShapinsky:
Moved dependencies to a separate docker image to test how that would effect build and testing times.


TShapinsky:
Recent CI on develop:
image
This PR:
image

PR #380, opened 2022-11-04
Merge 0.4.1 into Main

TShapinsky:


anyaelena (approved):

axelstudios:
In testing this release it looks like the `Remove Test Case` button in the UI no longer works?  Seems like a mismatch between the Mongo id and the `ref_id`.

The response:
 ```json
{
  "data": {
    "removeSite": null
  }
}
```

anyaelena (approved):

anyaelena (approved):
We need to add a release as well.

Issue #347, opened 2022-08-15
Cannot advance run from REST API


TShapinsky:
maybe `runs/{run_id}/advance`

PR #383, opened 2022-11-07
API Refactor and Improvements

axelstudios:
## PR Notes
- Refactored API to exclusively use mongo and redis
- Modularized API functionality with error handling
- Improved API documentation
- Added webpack caching for performance
- Added git SHA to `/api/v2/version` response
- Renamed `/models/` API endpoints to `/sites/`
- Refactored API handling so that Haystack and REST requests call the same function (in most cases)
- Faster `advance` response time
- Fixed all `nodehaystack` npm vulnerabilities
- Better invalid route handling

### Bugfixes
- Fixed site deletion
- Fixed bug where simulations were never saved and the _Completed-Simulations_ page was always empty
- Fixed schema mismatch between graphQL and the updated `simulation` collection

### Fixed Issues
#346
#347


alfalfa_web/server/alfalfa-server.js
```diff
@@ -169,6 +170,7 @@ class AlfalfaServer extends HServer {
     this.sites = this.db.collection("site");
     this.mrecs = this.db.collection("recs");
     this.recs = {};
+    this.api = new AlfalfaAPI(this.db, this.redis);
```
TShapinsky, line 173: I love this

alfalfa_web/server/api.js
```diff
@@ -0,0 +1,151 @@
...
+          realtime: String(!!realtime),
+          external_clock: String(!!externalClock)
+        }
+      };
+      const params = {
+        MessageBody: JSON.stringify(messageBody),
+        QueueUrl: process.env.JOB_QUEUE_URL,
+        MessageGroupId: "Alfalfa"
+      };
+      await this.sqs.sendMessage(params).promise();
```
TShapinsky, line 124: Should this flavor of job dispatching be moved to an API method?

alfalfa_web/server/advancer.js
```diff
@@ -32,20 +35,20 @@ class Advancer {
   }
 
   advance(siteRefs) {
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve) => {
```
TShapinsky, line 38: Possibly out of scope for this PR, but it would be nice to move most of the guts of this to an API function which takes `message_name`, and `params` and provides callbacks for when a response happens (error or otherwise). The message handling on the worker side is very standard so I don't see any reason to not have message send/receive not neatly packaged.

alfalfa_web/server/utils.js
```diff
@@ -8,6 +8,19 @@ function del(client, keys) {
...
 }
 
+// Return the value of a specific property from a redis hash
+function getHashValue(client, key, prop) {
+  return new Promise((resolve, reject) => {
+    client.hget(key, prop, (err, result) => (err ? reject(err) : resolve(result)));
+  });
+}
+
+function setHashValue(client, key, prop, value) {
```
TShapinsky, line 18: should these redis utilities be moved to the API?

.github/workflows/ci.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       - name: Build and run stack
         run: |
           printenv
-          docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build -d worker
+          docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build --build-arg GIT_COMMIT=$(git rev-parse --short=10 HEAD) -d worker
```
TShapinsky, line 81: Instead of using a bash command to get the commit hash you can use the `github.sha` from the github context https://docs.github.com/en/actions/learn-github-actions/contexts#github-context if that works for you

axelstudios: Perfect, that's what I was looking for.  Webpack worked fine locally with the `.git` directory, but then of course docker build doesn't have access to that..

alfalfa_web/server/advancer.js
axelstudios: I considered moving this to the API class, but since it already existed and was standalone I just leveraged it as-is

TShapinsky (approved):

PR #393, opened 2023-01-03
lock versions of dependent docker containers

nllong:
Our docker dependencies were not locked to specific versions, this fixes that.


docker-compose.yml
```diff
@@ -57,7 +57,7 @@ services:
       - MINIO_SECRET_KEY=${AWS_SECRET_ACCESS_KEY}
   mc:
     container_name: alfalfa_mc
-    image: minio/mc
+    image: minio/mc:RELEASE.2022-12-13T00-23-28Z
```
TShapinsky, line 60: wow, that's a version string if I've ever seen one.

TShapinsky (approved):
Looks good to me

PR #394, opened 2023-01-03
Upgrade dependencies

nllong:
Updates a bunch of core dependencies. This needs to be merged after #393 is merged to develop and develop is merged into this branch. Also, https://github.com/NREL/alfalfa-client/pull/47 needs to be merged and released as well.

Should also address #386, #389, and #391


alfalfa_worker/Dockerfile
```diff
@@ -173,19 +173,15 @@ RUN git clone --branch $PYFMI_TAG --depth 1 https://github.com/modelon-community
...
+RUN pip3 install poetry==1.3.1
 
 # Copy and install poetry-based dependencies
 COPY pyproject.toml $BUILD_DIR
 COPY poetry.lock $BUILD_DIR
 
-RUN poetry config virtualenvs.create false \
-    && poetry install --no-dev
+# Install 'production' dependencies from poetry
+RUN poetry install --no-dev
```
nllong, line 184: pretty big difference here now that we aren't disabling virtualenvs. Python that is used is now in a Poetry virtualenv. The "start-worker" script adds the virtualenv path to the system path to make everything work.

pyproject.toml
```diff
@@ -55,14 +50,16 @@ influxdb = "^5.3.1"
...
-autopep8 = "^1.4.4"
-pre-commit = "^2.20.0"
-pytest = "^6.2.5"
-pycodestyle = "^2.5.0"
-pytest-cov = "^3.0.0"
-pytest-timeout = "1.4.2"
-tox = "^3.21.4"
+# alfalfa-client = "^0.4.0"
+# alfalfa-client = {path = "../alfalfa-client"}
+alfalfa-client = { git = "https://github.com/nrel/alfalfa-client.git", branch = "upgrade-dependencies" }
```
nllong, line 55: comment this and uncomment the version'd release above, once `alfalfa-client` is released.

alfalfa_worker/Dockerfile
```diff
@@ -150,17 +149,25 @@ RUN git clone --branch $SUNDIALS_TAG --depth 1 https://github.com/LLNL/sundials.
     && make install \
     && cd .. && rm -rf sundials sundials_build
 
-# install and upgrade setuptools because this is needed for symfit (used in E+ Python)
-#    For Assimulo -- numpy, Cython must match the version in pypoetry.toml
-RUN ln -sf /usr/bin/python3 /usr/bin/python \
```
nllong, line 155: pretty big difference here now that we aren't disabling virtualenvs, so the python that is used is now in a Poetry virtualenv.

alfalfa_worker/Dockerfile
```diff
@@ -170,50 +177,39 @@ RUN git clone --branch $ASSIMULO_TAG --depth 1 https://github.com/modelon-commun
...
-RUN pip3 install setuptools==65.3.0
-
 WORKDIR /alfalfa
+
 # Include the path to where alfalfa_worker is so that system calls can find it.
 ENV  PYTHONPATH="/alfalfa:${PYTHONPATH}"
 
 COPY . /alfalfa
 COPY ./deploy/wait-for-it.sh /usr/local/wait-for-it.sh
-COPY ./deploy/start_worker.sh start_worker.sh
```
nllong, line 199: no need to copy this over, just call it from `/alfalfa/deploy/start_worker.sh`

alfalfa_worker/Dockerfile
```diff
@@ -150,17 +149,25 @@ RUN git clone --branch $SUNDIALS_TAG --depth 1 https://github.com/LLNL/sundials.
...
+RUN pip3 install poetry==1.3.1
+
+WORKDIR /alfalfa
+
+# Copy and install poetry-based dependencies
+COPY pyproject.toml /alfalfa
+COPY poetry.lock /alfalfa
+
+# Install 'production' dependencies from poetry
+RUN poetry install --only main
```
TShapinsky, line 165: The only issue I see with moving this here is that if any dependencies are changed we have to rebuild pyfmi and assimulo

alfalfa_worker/Dockerfile
```diff
@@ -170,50 +177,39 @@ RUN git clone --branch $ASSIMULO_TAG --depth 1 https://github.com/modelon-commun
...
+# update the path to python, which is in the Poetry virtual environment.
+# The path (mfknN3Ol) is generated by Poetry and should be updated if the
+# install path changes.
+#
+# The dependency warning still exists, but it is a warning. Eventually this
+# will be resolved when we move to a newer version of Ubuntu.
+# See https://bugs.launchpad.net/ubuntu/+source/python-debian/+bug/1926870
+# Error: RequestsDependencyWarning: urllib3 (1.26.13) or chardet (3.0.4) doesn't match a supported version!
+#        warnings.warn("urllib3 ({}) or chardet ({}) doesn't match a supported
+ENV PATH=/home/alfalfa/.cache/pypoetry/virtualenvs/alfalfa-mfknN3Ol-py3.8/bin:$PATH
```
TShapinsky, line 200: this is deterministic?

deploy/start_worker.sh
```diff
@@ -28,4 +28,5 @@ else
 fi
 
 cd /alfalfa
+echo "Python being used is: $(which python)"
```
TShapinsky, line 31: is this correct since the `star_worker.sh` script is not being run with poetry?

pyproject.toml
```diff
@@ -55,14 +47,16 @@ influxdb = "^5.3.1"
...
+autopep8 = "~2.0"
+pre-commit = "~2.21"
+pytest = "~7.2"
+pycodestyle = "~2.10"
+pytest-cov = "~4.0"
+pytest-timeout = "~2.1"
+tox = "~4.1"
```
TShapinsky, line 59: are we using tox for anything?

TShapinsky (changes requested):

alfalfa_worker/Dockerfile
nllong: I questioned this myself. It is deterministic as long as we keep the path /alfalfa. Crazy, right?!

alfalfa_worker/Dockerfile
nllong: I agree. But there was no way around this until we release our own pyfmi package and can just install it.

deploy/start_worker.sh
nllong: it is because we set the path to the python in the docker file

pyproject.toml
nllong: we aren't! I'll clean up

nllong:
@TShapinsky  -- removed the tox pieces. We need to do it in `alfalfa-client` too.

alfalfa_worker/Dockerfile
nllong:
```
=> [nrel/alfalfa-worker:latest base 20/25] RUN git clone --branch Assimulo-3.2.9 --depth 1 https://  124.3s
 => [nrel/alfalfa-worker:latest base 21/25] RUN git clone --branch PyFMI-2.9.5 --depth 1 https://git  176.8s
```
Looks like 5 minutes to rebuild when we change dependencies. But the poetry install in dev is faster now as it only installs the dev dependencies. Before we were creating two python virtual envs, one in the /home/build path and one in /alfalfa.

alfalfa_worker/Dockerfile
TShapinsky: huh

deploy/start_worker.sh
TShapinsky: makes sense

TShapinsky (approved):

alfalfa_worker/Dockerfile
nllong: Here is the poetry "generate_env_name" code for reference. hash created from current working directory.

https://github.com/python-poetry/poetry/blob/5f8b9da15dc13e44ac5bf3f2cd9ac1f72b3178ac/src/poetry/utils/env.py#L1147

PR #397, opened 2023-01-06
Retry if alfalfa gets bad response from e+

TShapinsky:
I've noticed this happening a couple times and not being handled well. So now it should retry and if that fails raise an error with meaningful information


nllong (approved):
interesting fix. looks good.

PR #401, opened 2023-01-20
Add Minimal Resstock OSW test

TShapinsky:
- [x] Fix bug where pyfmi would not be available when using `dev` compose file on my machine
- [x] Remove annual_run job (can add back if needed, but has never been used for anything)
- [x] Allow uploads of OSWs which do not have measures
- [x] Add minimal OSW with resstock model to test suite
- [x] Consolidate basic model running tests


anyaelena:
Was the annual_run for running non-interactive?

tests/integration/test_worker_scale.py
```diff
@@ -13,12 +13,7 @@
 
 
 @pytest.fixture
-def alfalfa_client():
-    yield AlfalfaClient(host='http://localhost')
-
-
-@pytest.fixture
-def models(alfalfa_client: AlfalfaClient):
+def scale_models(alfalfa: AlfalfaClient):
```
anyaelena, line 16: nice!

anyaelena (approved):

PR #423, opened 2023-03-17
Externalize Worker dependencies and improve CI speed

### LINKED ISSUE #414: build/publish alfalfa dependencies in external repo to save on build time.
see https://github.com/NREL/alfalfa/pull/378/files

TShapinsky:
Changes:
- worker, web, and grafana images are now pushed to github packages instead of dockerhub
- Docker builds use buildx and github actions cache to speed up image compilation between commits
- Worker base images has been migrated to [alfalfa-dependencies](https://github.com/NREL/alfalfa-dependencies)
- Energyplus startup errors now take a minimum of 5 seconds to fail instead of 2 minutes
- `test_basic_operations` has been slimmed
- Mass model compatibility checks have been moved to simulation tests since running without the client is much faster
- upgrade openstudio to v3.6.1
- upgrade e+ to 23.1.0
Fixes #414


alfalfa_worker/jobs/openstudio/lib/workflow/measures/alfalfa_python_environment/measure.rb
```diff
@@ -79,7 +79,7 @@ def run(ws, runner, user_args)
         python_paths.append(run_dir_match[1] + "/.venv/lib/python3.8/site-packages")
       end
 
-      i = 3
+      i = 4
```
anyaelena, line 82: what's up here?

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -35,9 +35,9 @@ def __init__(self, run_id, realtime, timescale, external_clock, start_datetime,
 
         # EnergyPlus MLEP initializations
         self.ep = mlep.MlepProcess()
-        self.ep.bcvtbDir = '/home/alfalfa/bcvtb/'
-        self.ep.env = {'BCVTB_HOME': '/home/alfalfa/bcvtb'}
-        self.ep.accept_timeout = 120000
+        self.ep.bcvtbDir = '/alfalfa/bcvtb/'
+        self.ep.env = {'BCVTB_HOME': '/alfalfa/bcvtb'}
+        self.ep.accept_timeout = 5000
```
anyaelena, line 40: We increased this timeout intentionally from one minute to two when working with a complex E+ model for BUILD-SOS.  Is there a reason you are confident that this initialization will only take 5 seconds now?

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -58,6 +58,7 @@ def __init__(self, run_id, realtime, timescale, external_clock, start_datetime,
         # it will be set to False once the desired start time is reach
         self.master_enable_bypass = True
         self.first_timestep = True
+        self.model_start_timeout = 300
```
anyaelena, line 61: why 300?

anyaelena (dismissed):
This is great!  I want to be sure you've considered implications of reducing that timeout but otherwise, good to go!

alfalfa_worker/jobs/openstudio/lib/workflow/measures/alfalfa_python_environment/measure.rb
TShapinsky: idf format changed and added an additional line of configuration to the python search path idf object.

alfalfa_worker/jobs/openstudio/step_run.py
TShapinsky: Yeah, so the change here is that instead of doing the accept once it will retry for `self.model_start_timeout` seconds. So while the `ep.accept_timeout` is only 5 seconds it will repeatedly retry for 5 minutes. This means that if the model has a catastrophic error and produces a `.err` file the user only has to wait 5 seconds before seeing the reason the model failed instead of the entire timeout period.

TShapinsky: Maybe I should add comments to those lines

anyaelena: love it!

anyaelena (approved):

PR #424, opened 2023-03-21
Temporarily remove automatic natural gas metering

TShapinsky:
Buildings which do not have natural gas equipment will error out if a sensor reading the natural gas meter is added.

In the future we could re add with some logic about when to add the natural gas meter.

### DIFF
--- a/alfalfa_worker/jobs/openstudio/lib/workflow/measures/alfalfa_site_sensors/measure.rb
+++ b/alfalfa_worker/jobs/openstudio/lib/workflow/measures/alfalfa_site_sensors/measure.rb
@@ -58,8 +58,7 @@ class AlfalfaSiteSensors < OpenStudio::Measure::EnergyPlusMeasure
     end
 
     fuels = [
-      'Electricity',
-      'NaturalGas'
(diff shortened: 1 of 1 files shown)


bonnema:
For posterity, a solution could be to check if any of the following objects use the fuel type `NaturalGas`
```
OtherEquipment
Exterior:FuelEquipment
HVACTemplate:System:VRF
HVACTemplate:Plant:Boiler
ZoneHVAC:HybridUnitaryHVAC
ZoneHVAC:HighTemperatureRadiant
Coil:Cooling:DX:CurveFit:Performance
Coil:Cooling:DX:MultiSpeed
Coil:Heating:Fuel
Coil:Heating:DX:MultiSpeed
AirConditioner:VariableRefrigerantFlow
Boiler:HotWater
Boiler:Steam
Chiller:EngineDriven
Chiller:CombustionTurbine
ChillerHeater:Absorption:DirectFired
WaterHeater:Mixed
WaterHeater:Stratified
Generator:InternalCombustionEngine
Generator:CombustionTurbine
Generator:MicroTurbine
```

PR #425, opened 2023-03-27
Bump versions to 0.5.0

TShapinsky:


axelstudios (approved):
Removed top-level `package.json`, updated version to `0.5.0` in alfalfa_web
- ESLint would definitely be a benefit, but it requires a good baseline config and can conflict with prettier.  It will need to be in the `alfalfa_web` scope

TShapinsky:
> Removed top-level `package.json`, updated version to `0.5.0` in alfalfa_web
> 
>     * ESLint would definitely be a benefit, but it requires a good baseline config and can conflict with prettier.  It will need to be in the `alfalfa_web` scope

oops, thanks for catching that. In recommitting after my `reset` I must've accidentally clicked the wrong `package.json`

PR #437, opened 2023-04-12
Report preprocessing errors and consolidate site information collection

TShapinsky:
Right now alfalfa will not show a site until after it has been processed and information has been extracted from the model. In cases where the model is somehow broken this results in an error being generated but not relayed to the user. This PR removed the requirement of a site entry to be generated in the database. I'm open to critiques of this technique. I do believe that we have a lot of redundancy between Site and Run collections at this current point in time, but now doesn't seem like the time to reconcile that.

### DIFF
--- a/alfalfa_web/server/api.js
+++ b/alfalfa_web/server/api.js
@@ -75,9 +75,10 @@ class AlfalfaAPI {
 
         if (site) {
           const siteHash = await getHash(this.redis, siteRef);
+          site = mapHaystack(site);
 
-          site_dict.name = site?.dis;
-          site_dict.datetime = siteHash?.sim_time;
+          site_dict.name = site?.dis ?? site_dict.name;
+          site_dict.datetime = siteHash?.sim_time ?? "";
         }
         return site_dict;
       }
--- a/alfalfa_web/server/api.js
+++ b/alfalfa_web/server/api.js
@@ -41,27 +41,11 @@ class AlfalfaAPI {
     try {
       const runs = [];
 
-      const models = (await this.models.find().toArray()).reduce(reduceById, {});
-      const sites = (await this.sites.find().toArray()).map(mapHaystack).reduce(reduceByRefId, {});
-
       for await (const run of this.runs.find()) {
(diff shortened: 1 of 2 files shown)


alfalfa_web/server/api.js
```diff
@@ -78,21 +62,24 @@ class AlfalfaAPI {
...
-            name: site?.dis,
-            status: run.status.toLowerCase(),
-            simType: site.sim_type,
-            datetime: siteHash?.sim_time ?? "",
-            uploadTimestamp: run.created,
-            uploadPath: `uploads/${model.ref_id}/${model.model_name}`,
-            errorLog: run.error_log
-          };
+          site_dict.name = site?.dis;
+          site_dict.datetime = siteHash?.sim_time;
```
axelstudios, line 80: Since this could set the value to undefined let's use `siteHash?.sim_time ?? ""`.  Same for _name_, if `model.model_name` was set that could be replaced with undefined if `site.dis` isn't set

alfalfa_web/server/api.js
```diff
@@ -78,21 +62,24 @@ class AlfalfaAPI {
...
+          status: run.status.toLowerCase(),
+          datetime: "",
+          simType: run.sim_type,
+          uploadTimestamp: run.created,
+          uploadPath: `uploads/${model.ref_id}/${model.model_name}`,
+          errorLog: run.error_log
+        };
+
         if (site) {
-          site = mapHaystack(site);
```
axelstudios, line 82: We probably want to keep the haystack mapper, right now the names have `s:` prefixes unless there's a preprocessing error: 
image

axelstudios (changes requested):

axelstudios:
Also, just noticed that the `Error` button text isn't aligned with the other status since the button has padding 😞 
image

alfalfa_web/server/api.js
TShapinsky: ahh, so that's why the haystack call was there.

axelstudios (approved):

PR #442, opened 2023-05-11
Add Schedule Generation Tests and Support for more Schedule Types

TShapinsky:
Adds `Schedule:Constant` support and creates tests validating generation and overrides.

Fixed an ems string escaping error which caused pre-pended `_` in cases where it shouldn't be


pyproject.toml
```diff
@@ -47,7 +47,8 @@ influxdb = "^5.3.1"
 # required for tests to pass right now. There are often issues when pinning to a branch
 # and calling poetry update. I have had to completely remove my venv and reinstall in order
 # to get the latest updates.
-alfalfa-client = "~0.5.0"
+alfalfa-client = { git = "https://github.com/nrel/alfalfa-client.git", branch = "handle_none_output_values" }
```
anyaelena, line 50: you meant to commit this?

tests/integration/models/schedule_model/modify_model.rb
```diff
@@ -0,0 +1,62 @@
+require 'openstudio'
+
+# puts OpenStudio::Model::Model.methods.filter { |m| m.to_s.upcase.include? "LOAD"}
+# exit
+model = OpenStudio::Model::Model.load("small_office.osm").get
```
anyaelena, line 5: Is it a design choice to make the model in code vs including and .os (maybe to keep OS version consistent), or am I missing something?

anyaelena (approved):

pyproject.toml
TShapinsky: yeah... That's actually this PR in the client which is required to run the tests https://github.com/NREL/alfalfa-client/pull/56

PR #458, opened 2023-07-06
Bump version to 0.6.0

TShapinsky:


anyaelena (approved):
Seems like ideally we would update version in only one place?

TShapinsky:
> Seems like ideally we would update version in only one place?

yeah, I guess the upside is that the version numbers in these files don't actually mean anything since we're not releasing either the web or worker as a package. The version that actually matters is the tag created for the release which the docker image gets tagged with.

PR #464, opened 2023-08-17
Convert site electricity sensor from Joules to watts

TShapinsky:
Add a program to divide site electricity sensor by 60 to convert it from Joules to Watts (Joules/Second).

### DIFF
--- a/alfalfa_worker/jobs/openstudio/lib/alfalfa-lib-gem/lib/utils.rb
+++ b/alfalfa_worker/jobs/openstudio/lib/alfalfa-lib-gem/lib/utils.rb
@@ -28,6 +28,7 @@ module OpenStudio
       # @return [String]
       def create_ems_str(id)
         id = id.gsub(/[\s-]/, '_')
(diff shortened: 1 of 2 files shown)


alfalfa_worker/jobs/openstudio/lib/alfalfa-lib-gem/lib/utils.rb
```diff
@@ -28,6 +28,7 @@ def get_idd_type(object)
       # @return [String]
       def create_ems_str(id)
         id = id.gsub(/[\s-]/, '_')
+        id = id.gsub(/[^\w]/, '_')
```
TShapinsky, line 31: Without this the function is not idempotent. It will append `_` forever

anyaelena:
@kbenne we had paused on merging this to consider whether the conversion should be handled in E+ instead.  Since the next Alfalfa release will remove MLEP, we aren't worried about lasting consequences.

anyaelena (approved):
see pr comment

PR #471, opened 2023-12-04
Begin alignment with new API

TShapinsky:
## API:
- New API endpoint structure
- Internal API functions now use Run and Point objects as parameters instead of the reference Id
- All mentions of site in the api have changed to run
- Aliases can replace runIds in the url (you no longer have to get the alias, you can just use it instead of the runId)
- Added error handler. Now API errors are passed through the API instead of being silenced.
- Run and Point objects are fetched by preprocessors to make handling consistent.
- Writing point values now validates points before committing writes.
- Response format is now consistent
- `sendRunMessage` api function will now reject a promise if it times out

## Worker:
- New OS measure to extract building name. Now works for E+ models as well.

## Model:
- Added metadata endpoint for single models
- Added download endpoint for single models

## Run:
- Now includes a name which defaults to the uploaded model.
- Deprecated site object

# TODO
- [ ] Add API tests outside the functionality of the `alfalfa-client`. There are a lot of endpoints that it doesn't use.
- [ ] Integrate new api documentation with old api documentation


alfalfa_web/components/Sites/ErrorDialog.js
```diff
@@ -2,20 +2,20 @@ import React from "react";
...
+            <span>{`${run.name} Error Log`}</span>
             <IconButton onClick={onClose}>
               <Close />
             </IconButton>
           </Grid>
         </DialogTitle>
         <DialogContent>
-          <pre style={{ whiteSpace: "pre-wrap" }}>{site.errorLog}</pre>
+          <pre style={{ whiteSpace: "pre-wrap" }}>{run.errorLog}</pre>
         </DialogContent>
```
TShapinsky, line 19: Changing from site to run

alfalfa_web/components/Sites/PointDialog.js
```diff
@@ -90,7 +100,7 @@ export const PointDialog = ({ onClose, site }) => {
       <Dialog fullWidth={true} maxWidth="lg" open={true} onClose={onClose}>
         <DialogTitle>
           <Grid container justifyContent="space-between" alignItems="center">
-            <span>{`${site.name} Points`}</span>
+            <span>{`${run.name} Points`}</span>
             <IconButton onClick={onClose}>
               <Close />
             </IconButton>
```
TShapinsky, line 106: changing from site to run

alfalfa_web/components/Sites/Sites.js
```diff
@@ -9,72 +9,72 @@ import { StartDialog } from "./StartDialog";
...
 
   const validStates = {
-    start: ["ready"],
-    stop: ["preprocessing", "starting", "started", "running", "stopping"],
-    remove: ["ready", "complete", "error"],
-    download: ["ready", "complete", "error"]
+    start: ["READY"],
+    stop: ["PREPROCESSING", "STARTING", "STARTED", "RUNNING", "STOPPING"],
+    remove: ["READY", "COMPLETE", "ERROR"],
+    download: ["READY", "COMPLETE", "ERROR"]
```
TShapinsky, line 21: I got rid of the code which was lowercasing all of the statuses when they went through the api

alfalfa_web/components/Sites/Sites.js
```diff
@@ -189,17 +189,13 @@ export const Sites = () => {
...
           <Grid item>
-            <Button
-              variant="contained"
-              disabled={isDownloadButtonDisabled()}
-              onClick={handleDownloadSite}
-              sx={{ m: 1 }}>
-              Download Site
+            <Button variant="contained" disabled={isDownloadButtonDisabled()} onClick={handleDownloadRun} sx={{ m: 1 }}>
+              Download Run
             </Button>
```
TShapinsky, line 199: Rest of file is migrating from sites to runs

alfalfa_web/server/api-v2.js
```diff
@@ -76,26 +87,119 @@ router.get(/\/$/, (req, res) => {
...
-    .catch(() => res.sendStatus(500));
+    .listRuns()
+    .then((payload) => res.json({ payload }))
+    .catch((err) => {
+      next(err);
+    });
+});
+
+router.param("runId", (req, res, next, id) => {
+  const error = validate(
```
TShapinsky, line 100: Whenever `runId` is a param in the URL this function will parse the id and add the `run` object to the request object. This allows all downstream function to not have to worry about validating and processing the id. This also allows me to check if an alias exists in cases where the id doesn't correspond to a run.

alfalfa_web/server/api-v2.js
```diff
@@ -76,26 +87,119 @@ router.get(/\/$/, (req, res) => {
...
+          res.status(404).json({ error: `Run with id '${id}' does not exist` });
+        } else {
+          res.sendStatus(500);
+        }
+      })
+      .catch(next);
+  }
+});
+
+router.param("pointId", (req, res, next, id) => {
```
TShapinsky, line 144: This is the same but just with `pointId` instead of `runId`

alfalfa_web/server/api-v2.js
```diff
@@ -76,26 +87,119 @@ router.get(/\/$/, (req, res) => {
...
+      } else if (point == null) {
+        res.status(404).json({ error: `Point with id '${id}' does not exist` });
+      } else {
+        res.sendStatus(500);
+      }
+    })
+    .catch(next);
+});
+
+router.param("modelId", (req, res, next, id) => {
```
TShapinsky, line 167: same again but for `modelId`

alfalfa_web/server/api-v2.js
```diff
@@ -110,47 +214,31 @@ router.get("/sites", async (req, res) => {
...
-
+router.get("/runs/:runId", async (req, res, next) => {
   api
-    .findSite(siteId)
-    .then((data) => {
-      if (data) {
-        return res.json({ data });
-      } else {
-        return res.status(404).json({
-          error: `Site ID '${siteId}' does not exist`
```
TShapinsky, line 137: This 404 has been removed as now that is handled by the pre-processor. You'll see this a lot in this file

alfalfa_web/server/api-v2.js
```diff
@@ -110,47 +214,31 @@ router.get("/sites", async (req, res) => {
...
-        });
-      }
+    .formatRun(req.run)
+    .then((payload) => {
+      res.status(200).json({ payload });
     })
-    .catch(() => res.sendStatus(500));
+    .catch(next);
```
TShapinsky, line 229: This catch will funnel all errors to the single error handler. You'll see this a lot in this file

alfalfa_web/server/api-v2.js
```diff
@@ -110,47 +214,31 @@ router.get("/sites", async (req, res) => {
...
-    .findSite(siteId)
-    .then((data) => {
-      if (data) {
-        return res.json({ data });
-      } else {
-        return res.status(404).json({
-          error: `Site ID '${siteId}' does not exist`
-        });
-      }
+    .formatRun(req.run)
```
TShapinsky, line 225: Requests to the internal api are now done with the run object instead of the id. This helps make everything more consistent and also allows us to change the id in the future without having to rewrite everything.

alfalfa_web/server/api-v2.js
```diff
@@ -164,362 +252,402 @@ router.get("/sites/:siteId", async (req, res) => {
...
- *               type: object
+ *               type: array
+ *               items:
+ *                 $ref: '#/components/schemas/pointMetadata'
+ */
+router.get("/runs/:runId/points", (req, res, next) => {
+  api
+    .getPointsByRun(req.run)
+    .then((points) => {
+      return points.map(api.formatPoint);
```
TShapinsky, line 285: All db objects now have an accompanying `format` function which is used to present them to the user.

alfalfa_web/server/api-v2.js
```diff
@@ -164,362 +252,402 @@ router.get("/sites/:siteId", async (req, res) => {
...
-    }
-  );
-  if (error) return res.status(400).json({ error });
-
+router.get("/runs/:runId/points/values", (req, res, next) => {
   api
-    .listPoints(siteId, /^(INPUT|BIDIRECTIONAL)$/)
-    .then((data) => res.json({ data }))
-    .catch(() => res.sendStatus(500));
+    .getPointsByRun(req.run)
```
TShapinsky, line 380: This should probably constrain to types output and bidirectional. Unless we want to allow users to query the current values of inputs. Which wouldn't be the worst thing.

alfalfa_web/server/api-v2.js
```diff
@@ -164,362 +252,402 @@ router.get("/sites/:siteId", async (req, res) => {
...
+            pointWrites.push([point, value]);
+          } else {
+            errors.push({ message: `Point with id '${pointId}' does not exist` });
+          }
+          return api.validatePointWrite(point, value).catch((err) => {
+            errors.push(JSON.parse(JSON.stringify(err, Object.getOwnPropertyNames(err))));
+          });
+        })
+        .catch((err) => {
+          errors.push(JSON.parse(JSON.stringify(err, Object.getOwnPropertyNames(err))));
```
TShapinsky, line 525: validate point writes and collate errors if there are ones.

alfalfa_web/server/api-v2.js
```diff
@@ -763,100 +843,115 @@ router.get("/sites/:siteId/download", (req, res) => {
...
+        .setAlias(alias, run)
+        .then(() => {
+          return res.sendStatus(204);
+        })
+        .catch(next);
+    })
+    .catch(next);
+});
 
+router.all("/aliases/:alias", (req, res, next) => {
```
TShapinsky, line 903: This does a similar thing as the other pre-processors. The reason this is an `all` instead of `param` is because that would fail on the `PUT` endpoint. Because when you `PUT` an alias it doesn't have to exist yet. So if you ever tried to create an alias you'd get a 404 because it didn't exist yet.

alfalfa_web/server/api.js
```diff
@@ -38,64 +37,53 @@ class AlfalfaAPI {
     });
   }
```
TShapinsky, line 38: Main thing in this file is making the api more consistent and also having functions operate on the db objects instead of their ids.

alfalfa_web/server/api.js
```diff
@@ -38,64 +37,53 @@ class AlfalfaAPI {
...
-        return site_dict;
-      }
-    } catch (e) {
-      console.error(e);
-      return Promise.reject();
+  getRun = async (runObjectId) => {
+    return this.runs.findOne({ _id: runObjectId });
+  };
+
+  formatRun = async (run) => {
```
TShapinsky, line 63: This was previously in two places, now it's a function.

alfalfa_web/server/api.js
```diff
@@ -104,306 +92,280 @@ class AlfalfaAPI {
...
-
-        // Delete runs
-        await this.runs.deleteMany({ ref_id: site.ref_id });
+  removeRun = async (run) => {
+    // Delete run
+    await this.run.deleteOne({ _id: run._id });
 
-        // Delete recs
-        await this.recs.deleteMany({ site: site._id });
+    if (run) {
```
TShapinsky, line 178: This if statement isn't necessary anymore.

alfalfa_web/server/api.js
```diff
@@ -104,306 +92,280 @@ class AlfalfaAPI {
...
   listModels = async () => {
-    try {
-      return (await this.models.find().toArray()).map(({ ref_id: id, model_name: modelName, created, modified }) => ({
-        id,
-        modelName,
-        created,
-        modified
-      }));
-    } catch (e) {
-      console.error(e);
```
TShapinsky, line 291: This kind of error handling was removed in favor of top level error handling to make the api more communicative about what went wrong.

alfalfa_web/server/api.js
```diff
@@ -104,306 +92,280 @@ class AlfalfaAPI {
...
-          { name: alias },
-          { $set: { name: alias, ref_id: refId } },
-          { upsert: true }
-        );
-      }
-    } catch (e) {
-      console.error(e);
-      return Promise.reject();
-    }
+  setAlias = async (alias, run) => {
```
TShapinsky, line 329: Aliases now use the `ObjectID` of a run instead of it's `ref_id` this is intended to make the look ups more performant as `ref_id` may not be indexed.

alfalfa_worker/jobs/openstudio/create_run.py
```diff
@@ -46,6 +46,13 @@ def exec(self):
         # run workflow
         check_output(['openstudio', '--gem_path', '/var/lib/gems/2.7.0', 'run', '-m', '-w', str(submitted_osw_path)])
 
+        self.logger.info('Loading Building Metadata')
```
TShapinsky, line 49: Runs now contain the name of the run. This is extracted by a measure which creates a report including this and possibly other data in the future.

anyaelena:
Closes #472

anyaelena (approved):
Thank you for turning this around so quickly!

anyaelena (approved):

PR #520, opened 2024-07-26
Multiplatform

TShapinsky:
- Upgrades to new version of `alfalfa-dependencies`
- Shrinks worker size from 4.8GB to 2.6GB
- Multiplatform builds of all components `linux/amd64` and `linux/arm64`
- Updates to Python 3.10 and Debian 12 (bookworm)

Size can be shrunk further by removing pre-installed python 3.8 dependencies which are just there to make the `refrig-case` test run faster. If we used this test less we could remove those as default installed dependencies and reduce the size by another ~400MB.


pyproject.toml
anyaelena, line 1: is this just cleanup of unnecessary dependencies post the move to model-specific python dependency files a while back?

.github/workflows/ci.yml
```diff
@@ -255,43 +255,46 @@ jobs:
...
 
       - name: Log into container registry
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Set up QEMU
```
anyaelena, line 264: can i get a quick update on what QEMU is doing?

anyaelena (approved):

TShapinsky:
QEMU is required in order to build the arm64 image. Eventually we might be
able to do the build on an arm64 worker, but for now it's emulated

pyproject.toml
TShapinsky: After moving to 3.10 we didn't need imporlib-metadata anymore and because the python installs are now split symfit moved to 3.8. I think upgrading to 3.10 also allowed the matplotlib pinning to be skipped. It's still installed though, just though a different dependency.

PR #541, opened 2025-01-06
Migrate to pyenergyplus

TShapinsky:
# Finished
- Added Unit to the output of a point metadata query.
- `LOGLEVEL` environment variable determines which logs are seen via stdout.
- Removed Haystack dependency from modelica.
- Remove BCVTB/ExternalInterface/mlep.
- Add Framework for created stepped simulation jobs which have external processes which want to control the main thread.
- Moved Job Exceptions into their own file.
- Remove GOAWS, replaced by redis.
- Upgraded dependencies (excluding influxdb package).
- Created test suite for different failure modes of pyeneregyplus to make sure cases are handled.
# In Progress
- Endpoint was added to serve raw logs (not just error logs) for a run.
  - Stream log handler created to push logs into redis.
  - `log_subprocess` function added to job which will automatically run a subprocess, pushing its output to the logger.
# Todo
- Clean up log messages to properly use log levels.
- Remove commented out code.
- Remove alfalfa ruby gem.
- Roll `alfalfa_metadata` measure into CreateRun job using `openstudio` module.
- Better code documentation for methods in `StepRunBase` and `StepRunProcess`
- Use redis stream handler for normal logs and error logs


alfalfa_web/server/api-v2.js
```diff
@@ -164,6 +164,13 @@ router.get("/runs/:runId/time", async (req, res, next) => {
     .catch(next);
 });
 
+router.get("/runs/:runId/log", async (req, res, next) => {
```
TShapinsky, line 167: This would allow job/run logs to be accessed from the web api.

alfalfa_worker/Dockerfile
```diff
@@ -21,10 +21,7 @@ ENV PYTHONPATH="${HOME}:${PYTHONPATH}"
 
 COPY ./alfalfa_worker ${HOME}/alfalfa_worker
 
-RUN pip3.8 install virtualenv \
-  && pip3.8 install \
-  scipy \
-  symfit
```
TShapinsky, line 27: The worker is no longer providing builtin python dependencies outside of what base alfalfa needs.

TShapinsky: This is partially due to removing the `refrig_case` test which required `symfit`. However `symfit` doesn't support python 3.12 and the project in general seems semi-abandoned so I removed the test case and replaced it with a small office based test model.

alfalfa_worker/__main__.py
```diff
@@ -3,14 +3,22 @@
...
 
 # Determine which worker to load based on the QUEUE.
 # This may be temporary for now, not sure on how else
 # to determine which worker gets launched
 from alfalfa_worker.dispatcher import Dispatcher
+from alfalfa_worker.lib.constants import DATETIME_FORMAT
 
 if __name__ == '__main__':
+
+    basicConfig(level=os.environ.get("LOGLEVEL", "INFO"),
```
TShapinsky, line 17: Output logs at `LOGLEVEL` and above to stdout.

alfalfa_worker/jobs/openstudio/__init__.py
```diff
@@ -2,5 +2,5 @@
 from pathlib import Path
 
 lib_dir = Path(os.path.dirname(__file__), 'lib')
-from .create_run import CreateRun
-from .step_run import StepRun
+# from .create_run import CreateRun
+# from .step_run import StepRun
```
TShapinsky, line 6: Removed to prevent source files from being automatically loaded. 

The `__init__.py` is run when the module gets imported, and in the dispatcher unit tests this can cause it to crash as `StepRun` has dependencies (`openstudio`, `pyenergyplus`) which may not be present on a developer's computer and are not present in the CI runner. 

This could be avoided by using a job other than `CreateRun` to confirm that the dispatcher is able to properly resolve job paths.

alfalfa_worker/jobs/openstudio/create_run.py
```diff
@@ -33,18 +31,18 @@ def exec(self):
         # If there are requirements.txt files in the model create a python virtual environment and install packaged there
         requirements = self.run.glob("**/requirements.txt")
         if len(requirements) > 0:
-            check_output(["python3.8", "-m", "venv", "--system-site-packages", "--symlinks", str(self.dir / '.venv')])
+            self.log_subprocess(["python3.12", "-m", "venv", "--system-site-packages", "--symlinks", str(self.dir / '.venv')])
```
TShapinsky, line 34: `log_subprocess` takes the output of the subprocess and logs it. This makes logs more useful as they include more of the output from the job.

alfalfa_worker/jobs/openstudio/lib/variables.py
```diff
@@ -1,140 +1,55 @@
...
 
 class Variables:
-    inputs: Dict[str, Dict]
-    outputs: Dict[str, Dict]
-    day_index: int = 0
-    hour_index: int = 1
-    minute_index: int = 2
-    month_index: int = 3
-
     def __init__(self, run: Run) -> None:
```
TShapinsky, line 9: Add better documentation to this class.

alfalfa_worker/jobs/step_run_base.py
```diff
@@ -1,15 +1,82 @@
...
+from alfalfa_worker.lib.utils import to_bool
+
+
+class ClockSource(AutoName):
+    INTERNAL = auto()
+    EXTERNAL = auto()
+
+
+@dataclass
+class Options:
```
TShapinsky, line 24: Consolidated options for `StepRunBase` extending jobs into one place.

TShapinsky: TODO: document each of these options better

alfalfa_worker/jobs/step_run_base.py
```diff
@@ -25,157 +92,79 @@ def __init__(self, run_id: str, realtime: bool, timescale: int, external_clock:
...
-        """Placeholder for getting write values from Mongo and writing into simulation BEFORE a simulation timestep"""
-
-    def write_outputs_to_redis(self):
-        """Placeholder for updating the current values exposed through Redis AFTER a simulation timestep"""
-
-    def update_sim_time_in_mongo(self):
-        """Placeholder for updating the datetime in Mongo to current simulation time"""
+        self.logger.info("Advancing to start time")
+        while self.run.sim_time < self.options.start_datetime:
+            self.logger.info("Calling advance")
```
TShapinsky, line 127: Make this a debug message, and increase the verbosity.

alfalfa_worker/lib/testcase.py
```diff
@@ -86,7 +86,7 @@ def __init__(self, **kwargs):
         self.u_store = copy.deepcopy(self.u)
         # Set default options
         self.options = self.fmu.simulate_options()
-        self.options['CVode_options']['rtol'] = 1e-6
+        # self.options['CVode_options']['rtol'] = 1e-6
```
TShapinsky, line 89: CVode no longer included by default. FMUs must provide solver.

docker-compose.dev.yml
```diff
@@ -16,32 +15,8 @@ services:
...
-      - REGION
-      - S3_BUCKET
-      - S3_URL
-      - S3_URL_EXTERNAL
-    depends_on:
-      - goaws
-      - mc
-      - mongo
-      - redis
-      - worker
```
TShapinsky, line 44: Removed redundant configuration.

tests/api/test_point.py
```diff
@@ -72,6 +73,7 @@ def test_point_retrieval(base_url, run_id, alfalfa_client):
     for point in outputs + bidirectionals:
         response = requests.get(f"{base_url}/runs/{run_id}/points/{point['id']}")
 
+        response.raise_for_status()
```
TShapinsky, line 76: this raises an exception if the return code isn't a 2XX.

tests/api/test_point.py
```diff
@@ -179,7 +182,7 @@ def test_point_writes(base_url, run_id):
     request_body = {
         'value': "hello"
     }
-    response = requests.put(f"{base_url}/runs/{run_id}/points/{inputs[0]['id']}", json=request_body)
+    response = requests.put(f"{base_url}/runs/{run_id}/points/{bidirectionals[0]['id']}", json=request_body)
```
TShapinsky, line 185: switched as default model no longer has any pure inputs. Only bidirectional points.

alfalfa_worker/lib/testcase.py
anyaelena: @TShapinsky @kbenne If we are changing expected format/requirements for FMU upload, we should update docs.

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -1,25 +1,37 @@
...
+def callback_wrapper(func):
+    def wrapped_func(self: "StepRun", state):
+        try:
+            return func(self, state)
+        except Exception:
+            self.catch_exception()
+    return wrapped_func
+
+
+class StepRun(StepRunProcess):
```
kbenne, line 29: Now the fun begins! 

I'm pretty sure I'm following the basic idea. Run EnergyPlus in a separate process and use signaling (`multiprocessing.Manager`) to coordinate between the main and the EnergyPlus process. Seems good to me, but makes me slightly sad that this is exactly what Spawn does and we haven't managed to consolidate this. Not your fault though, so what you have here is good. 

The main overarching comment I have is that the hierarchy of code seems a bit unnecessarily complex and maxes out my cognitive ability just a bit. I'm not suggesting a particular change, but just note that I believe the class hierarchy goes....

`StepRun` -> `StepRunProcess` -> `StepRunBase` -> `Job` and then there is `JobMetaClass` supporting `Job`. That's a lot or hierarchy to keep track of.

alfalfa_worker/jobs/openstudio/step_run.py
```diff
@@ -31,340 +43,304 @@ def __init__(self, run_id, realtime, timescale, external_clock, start_datetime,
...
-            self.check_error_log()
-            raise JobExceptionExternalProcess("EnergyPlus experienced initialization error")
-        elif _flag == -20:
-            self.check_error_log()
-            raise JobExceptionExternalProcess("EnergyPlus experienced time integration error")
-
-        self.ep.outputs = outputs
-
-    def step(self):
+    def start_simulation_process(self):
```
kbenne, line 58: and critically, this functions blocks while EnergyPlus is running so it is started in another Process. You might make a note about that.

alfalfa_worker/jobs/step_run_base.py
kbenne: Maybe that is an opportunity to consolidate the class hierarchy a little?

alfalfa_worker/lib/testcase.py
kbenne: Indeed FMUs for "co-simulation" are now the standard for BOPTEST and I guess here too and indeed that means the solver is baked into the FMU. 

I think this PR is key context.

https://github.com/ibpsa/project1-boptest/pull/536

alfalfa_worker/lib/run_manager.py
```diff
@@ -150,74 +148,6 @@ def checkout_run(self, run_id: str) -> Run:
         run.dir = run_path
         return run
 
-    def add_site_to_mongo(self, haystack_json: dict, run: Run):
```
kbenne, line 153: Is there any data left being pushed into a Mongo document? I don't see it. Can we remove Mongo entirely?

tests/integration/broken_models/small_office/bad_callback.osw
```diff
@@ -0,0 +1,19 @@
+{
```
kbenne, line 1: I admit that I'm running out of puff, but although I'm may not be catching all of the nuance on these tests, I'm really glad they are here. :)

kbenne (approved):
Hi @TShapinsky and @anyaelena. I left a lot of comments for thought, but this seems like a good PR and I'm excited to see it merged.

Nice job @TShapinsky

alfalfa_web/server/api.js
```diff
@@ -92,6 +92,11 @@ class AlfalfaAPI {
     return await getHashValue(this.redis, run.ref_id, "sim_time");
   };
 
+  getRunLog = async (run) => {
+    const log_lines = await this.redis.lRange(`run:${run.ref_id}:log`, -100, -1);
```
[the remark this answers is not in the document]
TShapinsky: Good question. `LRANGE` is O(S+N) where S is the start offset, and N is the number of elements. So it will currently grow linearly. However, if we are interested in getting only a few of the most recent lines we could switch from doing an `RPUSH` to `LPUSH`. That way we wouldn't need any offset to retrieve the log. This would make the writes and reads both constant complexity operations so there wouldn't be any issue with big logs, outside of storage space.

alfalfa_worker/Dockerfile
TShapinsky: Not quite. https://github.com/NREL/alfalfa/wiki/How-to-Migrate-EnergyPlus-Python-Plugins you can provide a `requirements.txt` which results in a virtual environment being created for your model which has your requirements installed.

alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -12,170 +11,128 @@ class StepRun(StepRunBase):
...
+
+            if id not in self.variables:
+                self.variables[id] = {}
 
-    def step(self):
+            if activate:
+                self.variables[id]["activate"] = input
+            else:
+                point = Point(ref_id=id, name=name, point_type=PointType.INPUT)
+                self.run.add_point(point)
```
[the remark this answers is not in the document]
TShapinsky: Point is common between all alfalfa jobs, whereas variables holds information specific to openstudio/energyplus. It is a bit of a vestigial design from when we needed to parse out the `variables.cfg` file for externalinterface. I may try to clean this up and consolidate the logic.

alfalfa_worker/jobs/modelica/step_run.py
```diff
@@ -12,170 +11,128 @@ class StepRun(StepRunBase):
...
+            activate = input.endswith("_activate")
+            name = input[0:-1 * (len("_activate"))] + "_u" if activate else input
+            id = name_to_id(name)
+
+            if id not in self.variables:
+                self.variables[id] = {}
 
-    def step(self):
+            if activate:
+                self.variables[id]["activate"] = input
```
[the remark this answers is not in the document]
TShapinsky: yeah, most of this is just moving around the way it was previously being done to get it to work with the new design and without haystack being baked into how everything was being done. IMHO the entire way we deal with modelica should be refactored, I don't like `Testcase` being separate from the job.

alfalfa_worker/jobs/openstudio/create_run.py
TShapinsky: That's a good point. I think this was an artifact from when I had multiple python versions installed at once.

alfalfa_worker/Dockerfile
anyaelena: Yes, this is staying the same, yes?  The E+ model with Python EMS + 3rd party Python modules just also needs to take responsibility for that requirements.txt.  This makes sense to me and I think it's important to continue to support this in Alfalfa.

alfalfa_worker/jobs/openstudio/create_run.py
anyaelena: Pls either make the update for this PR or add an issue.

PR #633, opened 2026-07-29
Fix apple silicon FMU support

nllong:
For Apple Silicon (M1/M2/M3+) users:

 Modelica FMUs only ship x86_64 (`binaries/linux64`) shared libraries, and pyfmi has no arm64 support, so the `worker` service is pinned to `platform: linux/amd64` in `docker-compose.yml` and runs under emulation on arm64 hosts (slower, but correct). 

This applies whether the worker image is pulled prebuilt from GHCR (which is published multi-arch, including a native arm64 variant that would otherwise be selected automatically) or built locally with `--build`. If you don't pin the platform (e.g. running the worker image directly with `docker run`, outside Compose), an arm64 worker will fail to load FMUs with a misleading error such as:


nllong:
Okay @TShapinsky  can you check this again? Alfalfa-dependencies were updated on that repo and I'm pulling from that GHCR now.

alfalfa_worker/Dockerfile
```diff
@@ -1,4 +1,13 @@
-FROM ghcr.io/natlabrockies/alfalfa-dependencies:prepare_080 AS base
+# The legacy SUNDIALS 5.x runtime libraries needed for Modelica-exported FMUs
+# (e.g. from Dymola/OpenModelica) that dynamically link against
+# libsundials_cvode.so.5 / libsundials_nvecserial.so.5 are built and
+# installed by the alfalfa-dependencies base image itself (see the
+# sundials-legacy-compat stage in that repo's Dockerfile), not here, so the
+# fix is built once and shared across every image built FROM this base
+# instead of being recompiled on every alfalfa_worker build. See the
+# "Alfalfa Dependencies" section of README.md for how the two repos/images
+# relate.
```
TShapinsky: this is unneeded

docker-compose.dev.historian.yml
```diff
@@ -0,0 +1,81 @@
+# Convenience overlay that combines development mode (live-reload web/worker)
```
TShapinsky: How is this different from the existing historian docker-compose?

docker-compose.dev.yml
```diff
@@ -3,7 +3,7 @@
 # container without needing to rebuild/restart Docker.
 #
 # Usage:
-#   docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build
+#   docker-compose -f docker-compose.yml -f docker-compose.dev.yml up --build --scale worker=2
```
TShapinsky, line 6: I'm not sure why this needed to be changed

docker-compose.dev.yml
```diff
@@ -25,6 +25,10 @@ services:
       - /srv/alfalfa/build
       - /srv/alfalfa/node_modules
   worker:
+    # Modelica FMUs only ship x86_64 (binaries/linux64) shared libraries, and pyfmi
+    # has no arm64 support, so always run the worker as amd64 (under emulation on
+    # arm64 hosts such as Apple Silicon) even though the image is published multi-arch.
+    platform: linux/amd64
```
TShapinsky: This can make the worker significantly slower, so it should not be the default state. it is also redundant if the platform is set in the main docker file as they are meant to be composed together.

README.md
```diff
@@ -14,6 +14,15 @@ We are currently working on increasing our developer documentation. See how to r
 
 Alfalfa runs as a Docker Compose stack (web, worker, MongoDB, Redis, and MinIO). Configuration is read from the `.env` file in the repository root. Requires Docker with the Compose plugin.
 
+**Note for Apple Silicon (M1/M2/M3+) users:** Modelica FMUs only ship x86_64 (`binaries/linux64`) shared libraries, and pyfmi has no arm64 support, so the `worker` service is pinned to `platform: linux/amd64` in `docker-compose.yml` and runs under emulation on arm64 hosts (slower, but correct). This applies whether the worker image is pulled prebuilt from GHCR (which is published multi-arch, including a native arm64 variant that would otherwise be selected automatically) or built locally with `--build`. If you don't pin the platform (e.g. running the worker image directly with `docker run`, outside Compose), an arm64 worker will fail to load FMUs with a misleading error such as:
+
```
TShapinsky, line 18: This is where it should mention that you can set the platform if needed.

docker-compose.yml
```diff
@@ -65,6 +65,10 @@ services:
     depends_on:
       - minio
   worker:
+    # Modelica FMUs only ship x86_64 (binaries/linux64) shared libraries, and pyfmi
+    # has no arm64 support, so always run the worker as amd64 (under emulation on
+    # arm64 hosts such as Apple Silicon) even though the image is published multi-arch.
+    platform: linux/amd64
```
TShapinsky: Remove.

TShapinsky (changes requested):

alfalfa_worker/Dockerfile
nllong: removed

docker-compose.dev.yml
nllong: the FMU example that is in the alfalfa notebooks requires two workers to be up so that the two models can interact. It really isn't needed here, but I figured it doesn't hurt to have two up for most users. The real issue is that the API doesn't have great handling of 'jobs' that are stuck because of not having enough workers.

docker-compose.dev.yml
nllong: understood. Yeah, I will move this to the comment section that you suggested.

README.md
nllong: updated a bit.

docker-compose.yml
nllong: removed.

docker-compose.dev.historian.yml
nllong: valid point. I deleted and made a comment in docker-compose.dev.yml on running devmode with historian as

```
docker-compose -f docker-compose.yml -f docker-compose.dev.yml -f docker-compose.historian.yml up --build --scale worker=2
```

docker-compose.elk.yml
```diff
@@ -3,7 +3,6 @@
 # Usage:
 #   docker-compose -f docker-compose.yml -f docker-compose.dev.yml -f docker-compose.elk.yml  up
 
-version: "3.4"
```
nllong, line 6: this is deprecated in docker compose now

docker-compose.historian.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.4"
```
nllong, line 1: this is deprecated in docker compose now

docker-compose.dev.yml
TShapinsky: let me know when this is addressed

nllong: Confirmed resolved — the `platform: linux/amd64` pin was removed from `docker-compose.dev.yml` in fbebc57 (right after this comment). I also noticed the README's Apple Silicon note still described the compose file pinning `platform: linux/amd64` automatically, which was stale since we removed that pin per your feedback here and in docker-compose.yml — updated it to reflect that users now opt in themselves via `DOCKER_DEFAULT_PLATFORM`/`--platform`.

TShapinsky:
@nllong I'm good to merge if you're done.

TShapinsky (approved):

